PR #1, opened 2022-01-10
Cleanz

Kirill888:
This is now importing ok, has some tests and should be passing linting with pylint/black/mypy.


.vscode/settings.json
```diff
@@ -0,0 +1,27 @@
...
+    "python.linting.pylintEnabled": true,
+    "python.sortImports.args": [
+        "--profile=black"
+    ],
+    "python.testing.pytestEnabled": true,
+    "python.testing.pytestArgs": [
+        "-s",
+        "-v"
+    ]
+}
```
gadomski: I generally don't love having editor-specific configuration in version control. Can we have these line-length arguments in one of the project configuration files?

gadomski (changes requested):

.vscode/settings.json
Kirill888: fair enough, I'll move that into a "developer" section of the readme.

`pylint` and `mypy` are configured via `pyproject.toml` file, but `flake8, pydocstyle, pycodestyle` are not because we don't run them on github actions. Also some of them might not support `pyproject.toml` config and would require some `.files` instead.

PR #2, opened 2022-01-12
Refactor: split overly large `_base.py` file

Kirill888:
What was a single `_base.py` is now

- `_crs.py` (pyproj CRS wrapper)
- `_geom.py` ("Shapely Geometry with CRS" class and helpers)
- `_geobox.py` (`GeoBox` class and related utilities)

This had to be done carefully to preserve git history across the split, hence two commits (step 1|2)

Checking history is preserved on all 3 new files:

- https://github.com/opendatacube/odc-geo/blame/refactor/odc/geo/_geobox.py
- https://github.com/opendatacube/odc-geo/blame/refactor/odc/geo/_geom.py
- https://github.com/opendatacube/odc-geo/blame/refactor/odc/geo/_crs.py

Note that there is a weak intentional cyclic dependency between `_crs.py` and
`_geom.py`. Mostly `_geom` needs CRS, that's a fundamental dependency. But CRS
also uses Geometry classes for some convenience methods, namely to report "valid
area" for a given CRS object.


gadomski (approved):
LGTM. Is there a compelling reason to keep these as "private" modules (i.e. with the leading `_`)? If there is a "public" API that we're committing to keep stable while changing the "internal" `_`-prefixed API, then it makes sense. I guess I'm curious if that is the plan.

Kirill888:
Right now almost everything is exposed from one flat namespace, hence the `_`. But we should think about having better hierarchy than just flat namespace.

PR #3, opened 2022-01-17
Some refactoring and more tests

Kirill888:
Mostly moving code around with the goal of adding type annotations without adding cyclic imports.

- Changing what is exposed in a flat namespace at `odc.geo.`, and what needs to be imported from otherwise not loaded modules.
   -  `GridSpec` and `GeoBox` are not pulled in by default
- Not importing `rasterio` in CRS module, assuming GDAL3 projection axis order always instead.
- Some more tests
- Splitting tools.py into `_roi.py` and  `_overlap.py`
- Fused `gbox.py` and `geobox.py` (preserving history)
- Split overly long `test_geometry`
- Moved all `xarray` tooling into one place


Kirill888:
This is getting too big. As it's mostly moving code around it's not really easy to review. So I'm gonna merge it before doing further changes that would benefit from not being hidden by all that code movement noise.

RichardScottOZ:
Makes sense to me.

PR #4, opened 2022-01-18
Reworking xarray interop

Kirill888:
- Rework CRS constructor (related to: https://github.com/opendatacube/datacube-core/issues/1222)
- moved testutils
- Using xarray extensions instead of just `.geobox`
- Using `encoding` instead of `attributes` for `grid_mapping` (https://github.com/opendatacube/datacube-core/issues/1084)
- adding tests

TODO:

add `.odc.` extension to Dataset as well.


odc/geo/_crs.py
```diff
@@ -16,11 +16,28 @@
...
+def _make_crs_key(crs_spec: Union[str, _CRS]) -> str:
+    if isinstance(crs_spec, str):
+        normed_epsg = crs_spec.upper()
+        if normed_epsg.startswith("EPSG:"):
+            return normed_epsg
+        return crs_spec
+    return crs_spec.to_wkt()
+
+
+@cachetools.cached(_crs_cache, key=_make_crs_key)
```
gadomski, line 31: Would it make sense to add a `lock` in case of multi-threaded access? Globals w/o a lock feel weird to me but mabye that's just me.

gadomski (approved):

odc/geo/_crs.py
Kirill888: I'm not sure how cachetools deals with concurrency actually. Dictionaries are "thread-safe" in Python, and so are imports, so wrapping cache around a function should be thread-safe as it happens at import time, and then I assume cachetools does reasonable thing afterwards, which at worst will create 2 copies of the same CRS, but will only cache one.

Kirill888: There is an option to supply `lock` to `@cached` decorator, also should probably replace plain `Dict` with some cache class instead.
more details here: https://cachetools.readthedocs.io/en/stable/#cachetools.cached

given that this is not part of the API, I think we can delay this change until later. Also need to check `pyproj` constraints for multi-threaded access.

PR #6, opened 2022-01-19
More import renames

Kirill888:
- `odc.geo.` is now fairly sparse with only classes and typedefs
  - probably worthwhile importing `GeoBox` in there too, now that it doesn't pull in `xarray`
- Add `odc.geo.xr` that hooks into `.odc` property on xarray objects when imported
- De-duplicating some common code between array/dataset extensions
- Cleaning up tests
  - removing old references to `geometry` module
  - some pylint guided fixes
- Packaging
  - Make `xarray` and `rasterio` dependencies optional


tests/test_geom.py
```diff
@@ -734,21 +731,19 @@ def test_3d_point_converted_to_2d_point():
...
     assert len(p_3d.coords[0]) == 2
 
     assert p_2d == p_3d
 
 
 def test_mid_longitude():
-    assert geometry.mid_longitude(geometry.point(10, 3, "epsg:4326")) == 10
-    assert geometry.mid_longitude(geometry.point(10, 3, "epsg:4326").buffer(3)) == 10
+    assert geom.mid_longitude(geom.point(10, 3, "epsg:4326")) == 10
+    assert geom.mid_longitude(geom.point(10, 3, "epsg:4326").buffer(3)) == 10
```
gadomski: When I ran tests locally, this failed:

```
FAILED tests/test_geom.py::test_mid_longitude - AssertionError: assert 10.00000000000001 == 10
```

gadomski (changes requested):

tests/test_geom.py
Kirill888: yeah, surprising it worked, added approx

PR #7, opened 2022-01-25
Be more like rioxarray in CF attribute handling

Kirill888:
netcdf/cf related changed:

- For CRSs supported by netdcf/cf convention include all attributes as reported by `pyproj.to_cf()` method.
- Look in `crs_wkt` attribute as well as in `spatial_ref` when looking for and interpreting CRS coordinate.

Some further API tweaking
- `with_crs="name"|True|False` argument to `xr_coord` was confusing, so it was replaced with
- `crs_coord_name="name"|None` where `None` is the same `False` was before, i.e. do not create a separate CRS coordinate.
- use default WKT format in `.wkt` property of the CRS object

### DIFF
--- a/odc/geo/_xr_interop.py
+++ b/odc/geo/_xr_interop.py
@@ -183,15 +183,14 @@ def assign_crs(
 
 
 def xr_coords(
-    gbox: GeoBox, with_crs: Union[bool, str] = True
+    gbox: GeoBox, crs_coord_name: Optional[str] = "spatial_ref"
 ) -> Dict[Hashable, xarray.DataArray]:
     """
     Dictionary of Coordinates in xarray format.
 
-    :param with_crs:
-       By default includes netcdf/cf style CRS Coordinate
-       with default name 'spatial_ref', if with_crs is a string then treat
-       the string as a name of the coordinate. To disable pass in ``False``.
+    :param crs_coord_name:
+       Use custom name for CRS coordinate, default is "spatial_ref".
+       Set to ``None`` to not generate CRS coordinate at all.
 
     Returns
     =======
@@ -201,11 +200,6 @@ def xr_coords(
     where names are either ``y,x`` for projected or ``latitude, longitude`` for geographic.
 
     """
-    spatial_ref = "spatial_ref"
-    if isinstance(with_crs, str):
-        spatial_ref = with_crs
-        with_crs = True
-
     attrs = {}
     crs = gbox.crs
     if crs is not None:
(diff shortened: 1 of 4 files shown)


odc/geo/_xr_interop.py
```diff
@@ -123,21 +123,16 @@ def spatial_dims(
...
-    if crs.projected:
-        grid_mapping_name = crs._crs.to_cf().get("grid_mapping_name")
-        if grid_mapping_name is None:
-            grid_mapping_name = "??"
-        grid_mapping_name = grid_mapping_name.lower()
-    else:
-        grid_mapping_name = "latitude_longitude"
-
+    cf = crs.proj.to_cf()
+    cf.setdefault("grid_mapping_name", "??")
```
gadomski: Is there a reason `??` is used? Should the be a constant somewhere? Or is this just for printing purposes?

odc/geo/_xr_interop.py
```diff
@@ -123,21 +123,16 @@ def spatial_dims(
...
     epsg = 0 if crs.epsg is None else crs.epsg
+    crs_wkt = cf.get("crs_wkt", None) or crs.wkt
```
gadomski, line 128:
```suggestion
    crs_wkt = cf.get("crs_wkt", crs.wkt)
```

gadomski (changes requested):

odc/geo/_xr_interop.py
Kirill888: Reason why I used `or` syntax is because it short-circuits, so we don't need to recompute `wkt` string if it is already included in the cf convention output (it always is, as far as I can tell). It's probably not a very expensive computation, but there is also no telling how often this code might be used in an application.

odc/geo/_xr_interop.py
Kirill888: Actually I'm not sure we should be setting it to anything. Basically `grid_mapping_name` is an `enum` defined by CF convention, so if your CRS is not supported by CF convention, it should probably not have `grid_mapping_name` defined at all. We are relying on it being there do detect "CRS coordinates" attached to an xarray. But this should probably change to look for any of these attributes instead: `grid_mapping_name, spatial_ref, crs_wkt`.

Not sure what that means for compatibility of files produced with `.to_netcdf(...)`. 


https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/apf.html

Kirill888: hm, looks like we are not even using that attribute for locating crs coordinate. I think it should just be left unset if it's not defined in the output of `.to_cf()`.

I should also add tests for round-tripping CRS to netcdf/zarr/tiff. it could be that non-standard name is worse than no name.

PR #8, opened 2022-01-26
Starting documentation building and cleanup

Kirill888:
This is far from finished...

- Enable documentation building 
- For now just get API docs sorted
- Cleaning up obvious problems
- Still not sure what is the best way to document classes, there is `autoclass` for example, but not sure how to combine that with `autosummary`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -268,3 +268,61 @@ jobs:
 
         env:
           DASK_TEMPORARY_DIRECTORY: /tmp/dask
+
+  check-docs:
+    runs-on: ubuntu-latest
+
+    needs:
+      - build-test-env-base
+      - run-black-check
+
+    steps:
+    - uses: actions/checkout@v2
+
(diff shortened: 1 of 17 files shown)


gadomski (approved):
Looks fine, only two comments:

- I know you're generally using conda, but since the repo works fine w/ pip, it might be nice to have a `requirements-dev.txt` or something similar to allow quick installation of dev/doc dependencies on a local system.
- I got `WARNING: Cannot resolve forward reference in type annotations of "odc.geo.crs.CRS.valid_region": name 'geom' is not defined` when building the docs locally. Wasn't sure a quick fix since it's a circular import problem...

Kirill888:
yeah that warning is because of this guard:

https://github.com/opendatacube/odc-geo/blob/ffd4a3880c48315e981d7864793615b597db0fe0/odc/geo/crs.py#L309-L310

I haven't tried to resolve it yet.

PR #10, opened 2022-02-01
Refactor gridspec

Kirill888:
`GridSpec` is a handy class for large scale processing, but it's API in datacube had some issues. Fixing those. It does break backwards compatibility with what was in datacube, but I think it was broken enough not to keep it that way.

1. Use pixel units to define grid, not CRS units

Using tile size in CRS units + resolution leads to a possibility of fractional
pixel size. If tile size is 7m and pixel size is 3m, what should the size of a
tile be in pixels? Then there are problem of numeric precision. Defining grid in
pixels using integer sizes doesn't have this problem.

2. Negative tile sizes not allowed

Previously one could use negative tile size to reverse the order for grid
indexes. This is confusing and was not documented well.

Instead `fplix/flipy` parameters are introduced.

3. Adding more convenient construction options
   - From some tile geometry, index and shape
   - Direct support for slippy tiles regime

4. Adding geojson output for debugging and visualization
5. `origin` parameter is a point in CRS units, so changed order there to `x, y` rather than `y, x`, in `datacube` that parameter was not really used, and always defaulted to `0,0`.

Also adding geojson output to geometry and and gridspec classes, was needed for debugging and is generally useful.

### About X/Y, Y/X confusions

I almost started addressing it in this PR, but decided that won't be so great, but I still think a better solution is needed. Basically there are a lot of places where a tuple of two numbers is used to configure one of these things:

- Size in pixels (usually Y,X order, like numpy `.shape`)
- Size of a pixel (`resolution`), in Y,X order to match `.shape`, but it's way less obvious choice and X,Y could just as well be assumed by the user. There is also negative resolution for Y to indicate that pixel row index change in the opposite direction of pixel location along Y axis.
- Point in CRS space, usually in X, Y order.
- 2-d index, usually in Y,X order to match `ndarray` element access order

I think it's worthwhile refactoring those plain tuples out of the API entirely. It should be obvious at construction point what order you want to specify, and also at use site. I'm thinking something along these lines

```python
def some_method(..., resolution: Resolution):
    rr = resolution
    rx, ry = rr.xy
    ry, rx = rr.yx
    print(f"x: {rr.x}, y: {rr.y}")

some_method(..., resolution = xy_(10, -10))
some_method(..., resolution = yx_(-10, 10))
some_method(..., resolution = res_(x=10, y=-10))
```

idea is to decouple construction order from use order, while still allowing use of tuples on both sides but with clear demarcation of the order. Obviously having named access/construction should also be supported.

There should be a generic container for any two X,Y things of the same type, but also types for Shape/Point/Index/Size/Resolution.

### DIFF
--- a/odc/geo/geom.py
+++ b/odc/geo/geom.py
@@ -466,6 +466,43 @@ class Geometry:
 
         return geom._to_crs(crs)
 
+    def geojson(
+        self,
+        properties: Optional[Dict[str, Any]] = None,
+        simplify: float = 0.05,
+        resolution: Optional[float] = float("+inf"),
+        wrapdateline: bool = False,
+        **props,
+    ) -> Dict[str, Any]:
+        """
+        Render geometry to GeoJSON.
+
+        Convert geometry to ``ESPG:4326`` and wrap it in GeoJSON Feature with supplied properties.
+
+        :param properties:
+            Properties to include in the GeoJSON output.
+
+        :param simplify:
+            Tolerance in degrees for simplifying geometry after changing to lon/lat. Larger number
+            will result in a smaller (fewer points) and hence faster to display, but less precise
+            geometry. Default is ``0.05`` of a degree. To disable set to ``0``.
+
+        :param resolution:
+           When supplied, extra points will be added to the original geometry such that no segment
+           is longer than ``resolution`` units. Passed on to
+           :py:meth:`~odc.geo.geom.Geometry.to_crs`.
+
+        :param wrapdateline: Passed on to :py:meth:`~odc.geo.geom.Geometry.to_crs`
+
+        :return: GeoJSON Feature dictionary
+        """
+        gg = self.to_crs("epsg:4326", resolution=resolution, wrapdateline=wrapdateline)
+        if simplify > 0:
+            gg = gg.simplify(simplify)
+        if properties is None:
+            properties = dict(**props)
+        return {"type": "Feature", "geometry": gg.json, "properties": properties}
+
     def split(self, splitter: "Geometry") -> Iterable["Geometry"]:
         """shapely.ops.split"""
         if splitter.crs != self.crs:
--- a/odc/geo/gridspec.py
+++ b/odc/geo/gridspec.py
@@ -4,50 +4,74 @@
 # SPDX-License-Identifier: Apache-2.0
 """GridSpec class."""
 import math
-from typing import Iterator, Optional, Tuple
+from typing import Any, Dict, Iterator, Optional, Tuple, Union
 
 from affine import Affine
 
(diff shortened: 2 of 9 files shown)


odc/geo/geom.py
```diff
@@ -466,6 +466,43 @@ def to_crs(
...
+
+        :return: GeoJSON Feature dictionary
+        """
+        gg = self.to_crs("epsg:4326", resolution=resolution, wrapdateline=wrapdateline)
+        if simplify > 0:
+            gg = gg.simplify(simplify)
+        if properties is None:
+            properties = dict(**props)
+        return {"type": "Feature", "geometry": gg.json, "properties": properties}
+
```
gadomski, line 505: What are your thoughts on using https://pypi.org/project/geojson/? If this is the only time this repository touches GeoJSON then it's probably overkill, but if there's other places you work w/ it maybe the external dependency is worth it?

odc/geo/gridspec.py
```diff
@@ -4,50 +4,74 @@
...
+    :param tile_shape:
+       ``(Y, X)`` size of each tile in pixels
+    :param resolution:
+       ``(Y, X)`` size of each data point in the grid, in CRS units. ``Y`` will usually be negative
+    :param origin:
+       ``(X, Y)`` coordinate of a bottom-left corner of the ``(0,0)`` tile in CRS units. Default is
+       ``(0.0, 0.0)``
+    :param flipx: when ``True`` grid index for X axis increments left to right
+    :param flipy: when ``True`` grid index for Y axis increments top to bottom
     """
```
gadomski, line 33: I don't see it, but maybe I missed it -- is there a `to_transform()` method (or something like it) to get a [GDAL transform](https://gdal.org/tutorials/geotransforms_tut.html) from a `GridSpec`?

gadomski (approved):

odc/geo/gridspec.py
Kirill888: There is none, because `GridSpec` is an "infinite" plane of tiles, there is a `transform` on each tile accessible via `GeoBox` class. So I guess you can say that `gs[0,0].geobox.transform` is a kind of "whole plane" transform, but not sure what flipping of axis does to that concept.

Think of grid spec more like an axis aligned 2-d binning abstraction, where each bin is an axis aligned geo registered bounded plane.

odc/geo/geom.py
Kirill888: I think it mostly deals with various geometry classes, but we get that from shapely so not sure whether it will make it easier or require extra conversion methods. Maybe for things like encoding native CRS, but even those aren't that hard. So I think it mostly duplicates shapely functionality, but I'll have a look.

PR #11, opened 2022-02-04
XY refactor

Kirill888:
There were a bunch of places where it was not clear whether some tuple contains X,Y or Y,X data. It's confusing on the interface and confusing when reading code. This PR adds an `XY[T]` type and a bunch of constructor methods that make it easy to understand what order x,y parameters are added.

Instead of using plain tuples those new types are used instead. On the input `shape=` parameters are still allowed to be supplied as plain `(int, int)` tuples and are equivalent to `iyx_(y, x)`. Internally those properties are kept in `XY[int]` form. Similarly a 2d index is also allowed to be supplied as `(int, int)`, as this is what allows `some_thing[ix, iy]` or `some_thing[row, col]` syntax, order depends on use case, but should be clearly stated in the docs.

Resolution input parameters are allowed to be a single float, but plain tuples are not allowed as this causes order confusion. Single value resolution is equivalent to `resyx_(-value, value)`.

## Change GeoBox constructor to accept shape

`Geobox(width, height, ...)` interface is problematic for several
reasons:

1. not compatible with `ndarray`, where shape is used instead
2. no type-safety, easy to get order wrong
3. two parameters to configure single property

Replacing `width, height` with a single parameter `shape`. This could be a simple `(int, int)` tuple for easier interop with `ndarray`. But one can choose to use more typesafe `XY[int]`.

Example of the change at call site

```python
   # before change:
   GeoBox(512, 256, A, "espg:4326")

   # becomes
   ## if array order is "native" at call-site
   GeoBox((256, 512), A, "epsg:4326")
   ## if width/height orded is preferred
   GeoBox(ixy_(512, 256), A, "epsg:4326")
```


tests/test_geobox.py
```diff
@@ -145,36 +147,37 @@ def test_geobox():
 
         assert GeoBox.from_geopolygon(polygon, resolution, crs=geobox.crs) == geobox
 
-        assert abs(resolution[0]) > abs(
+        # check that extra padding added by alignment is smaller than pixel size
+        assert abs(resolution.x) > abs(
```
Kirill888, line 151: note that this changes the test, it was incorrectly using Y resolution to check for padding change in X axis, but since both resolutions are of the same magnitude it didn't cause problems. So it's not a very good test, need a test with non-square pixels to check padding correctness.

odc/geo/geobox.py
```diff
@@ -185,16 +203,16 @@ def dimensions(self) -> Tuple[str, str]:
...
+    def resolution(self) -> Resolution:
         """Resolution in Y,X dimension order."""
-        return self.affine.e, self.affine.a
+        rx, _, _, _, ry, *_ = self.affine
+        return resxy_(rx, ry)
 
     @property
-    def alignment(self) -> Tuple[float, float]:
+    def alignment(self) -> XY[float]:
         """Alignment of pixel boundaries in Y,X dimension order."""
```
gadomski: Documentation is out of line with signature.

odc/geo/geobox.py
```diff
@@ -185,16 +203,16 @@ def dimensions(self) -> Tuple[str, str]:
         return crs.dimensions
 
     @property
-    def resolution(self) -> Tuple[float, float]:
+    def resolution(self) -> Resolution:
         """Resolution in Y,X dimension order."""
```
gadomski: `Resolution` looks like it's X,Y here: https://github.com/opendatacube/odc-geo/pull/11/files#diff-475a79dc4384c54f2803075a878e770e303b640df0d6c57c922c4ecec7b5c33fR120. If so, documentation is out-of-date.

odc/geo/gridspec.py
```diff
@@ -82,13 +95,15 @@ def dimensions(self) -> Tuple[str, str]:
...
         """Pixel boundary alignment."""
-        y, x = (orig % abs(res) for orig, res in zip(self.origin, self.resolution))
+        y, x = (
+            orig % abs(res) for orig, res in zip(self.origin.yx, self.resolution.yx)
+        )
         return (y, x)
 
     @property
     def tile_shape(self) -> Tuple[int, int]:
         """Tile shape in pixels (Y,X order, like numpy)."""
```
gadomski: Would it make sense to change the return signature to `YX` to make it clearer to e.g. type hinting tools?

gadomski (changes requested):

odc/geo/gridspec.py
Kirill888: Maybe? kinda on the fence about `.shape`, I feel like for `.shape` there is no ambiguity, so adding extra indirection for interfacing to ndarray might be annoying/confusing. Maybe it's worth creating `Shape(XY[int])` and implementing `__iter__` on that so that you could pass it in to `numpy` functions where `shape=` is expected, while still being able to use all the other `XY` methods too.

odc/geo/geobox.py
Kirill888: That's a doc comment issue, `Resolution` is basically `XY[int]`, so there is no "dimension order".

odc/geo/gridspec.py
```diff
@@ -22,46 +34,47 @@ class GridSpec:
...
 
         self.crs = norm_crs_or_error(crs)
         self._shape = tile_shape
         self.resolution = resolution
-        self.tile_size = (
-            tile_shape[0] * abs(resolution[0]),
-            tile_shape[1] * abs(resolution[1]),
+        self.tile_size = xy_(
+            tile_shape.x * abs(resolution.x),
+            tile_shape.x * abs(resolution.y),
```
Kirill888:
```suggestion
            tile_shape.y * abs(resolution.y),
```

clearly tests are not checking non-square tiles

odc/geo/geobox.py
Kirill888:
```suggestion
    def alignment(self) -> XY[float]:
        """Alignment of pixel boundaries in CRS units."""
```

odc/geo/geobox.py
Kirill888:
```suggestion
    def resolution(self) -> Resolution:
        """Resolution, pixel size in CRS units."""
```

Kirill888:
@gadomski

since your last review:

- fixed doc strings that still contained order where it no longer made sense
- fixed error in `GridSpec` and added test that would have caught it
- Added `Shape2d(XY[int])` class
  - it behaves just like a `Tuple[int, int]` would: can be unpacked, indexed, iterated, is a sequence, can be concatenated with other tuples on both sides, can be passed into `np.zeros(shape=<>, ...)`.
  - So it's not safe on "use-site" unless you choose to use it in a "safe" way, but you can use it with use-site readable order since it is `XY[int]`
  - The default order for the unsafe use is Y,X to match numpy
- Switched over (most?) interfaces to `Shape2d`
- Adding new methods/classes to api docs
- Fixed docs building warning and enabled strict doc building mode.

gadomski (approved):

PR #15, opened 2022-02-17
Cleanup math py and some refactor in GeoBox

Kirill888:
- Add math.py to docs and cleanup doc comments
- Move some standalone geobox functions into the class itself, those should have been methods from the start really

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -68,17 +68,18 @@ odc.geo.geobox
    GeoBox
    GeoboxTiles
 
-   affine_transform_pix
    bounding_box_in_pixel_domain
+   geobox_intersection_conservative
+   geobox_union_conservative
+   scaled_down_geobox
+
+   affine_transform_pix
    flipx
    flipy
    gbox_boundary
-   geobox_intersection_conservative
-   geobox_union_conservative
    pad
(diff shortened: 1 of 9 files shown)


gadomski (approved):
No issues from my side, but I'm curious if some of the work being done in `math.py` could be done via numpy rather than doing manually? Are there performance reasons behind the decision to implement e.g. snapping directly?

Kirill888:
> No issues from my side, but I'm curious if some of the work being done in `math.py` could be done via numpy rather than doing manually? Are there performance reasons behind the decision to implement e.g. snapping directly?

I don't think `numpy` has these things, but it's huge so maybe? These are mostly used for three things

1. Separating pixel aligned + sub-pixel translation, so that sub-pixel part can be ignored if under threshold
2. Rounding to nearest integer and actually converting to integer type but only if within some tolerance
3. Finding nearest width/height that is a multiple of some number, usually 16 or 8 because of some image formats constraints

I don't think these parts are performance sensitive, but if it were used in a tight loop crossing numpy boundary for just 1 or 2 values might be costly, but at this point you are better off figuring out how to perform whole computation on a larger numpy array rather than in a loop.

PR #21, opened 2022-03-21
ENH: Use reference to extras in setup.cfg all section

snowman2:


Kirill888:
That's cool, didn't know that was a thing

PR #23, opened 2022-03-25
Preparing for rc0

Kirill888:
## Code

- Adding `GeoBox` construction from bounding box
- Adding visualization things, svg output, html repr output
- Adding various misc convenience bits that were missing and were needed when developing visualization parts
- Adding geometry transform with affine matrix
- Adding data folder with geojson of oceans and css for geobox html display
- Minor fix in geobox slicing when single pixel row/column is selected

## Docs

- better overview
- update conda install instructions


setup.cfg
```diff
@@ -1,5 +1,5 @@
 [metadata]
-name = odc-geo
+name = odc_geo
```
gadomski: Why the name change?

Kirill888: warning from `pip` in python the two are equivalent anyway.

gadomski: From which command? I just tried `pip install -e .` and `python -m build` and neither gave me a warning. IMO it feels better to have the name match the github slug?

Kirill888: actually, I'm not longer sure if I was reading the right warning, I was failing to include data into packaging at the time, let's try it again with the previous name.

Kirill888: @gadomski warning was coming from `options.package_data` not from package name, I have reverted package name change and fixed package_data section.

gadomski (approved):

PR #35, opened 2022-04-20
Some more usage driven improvements

Kirill888:
- Refactor in `GeoboxTiles` to allow access to "roi" of a tile not just geobox, needed for data loading logic in odc-stac
  - `odc.roi.Tiles` now contains logic for that, independent of geobox
- Add convenience methods `GeoBox.{left,right,top,bottom}` for moving geobox
- Add `compute_output_geobox()` and expose it as `xx.odc.output_geobox()` as well
  - Construct appropriat geobox for projecting to some other reference (figures out resolution and bounds and makes GeoBox)
- Use `compute_output_geobox` in `xx.odc.reproject(..)` when input is CRS and not fully specified geobox
- Fixed errors in display of empty/single line geoboxes
- more precise type info in `crs.py`

### DIFF
--- a/odc/geo/geobox.py
+++ b/odc/geo/geobox.py
@@ -24,6 +24,7 @@ from .geom import (
     polygon_from_transform,
 )
 from .math import clamp, is_affine_st, is_almost_int
+from .roi import Tiles as RoiTiles
 from .roi import align_up, polygon_path, roi_normalise, roi_shape
 from .types import (
     XY,
@@ -913,13 +914,8 @@ class GeoboxTiles:
         :param box: source :py:class:`~odc.geo.GeoBox`
         :param tile_shape: Shape of sub-tiles in pixels ``(rows, cols)``
         """
-        tile_shape = shape_(tile_shape)
         self._gbox = box
-        self._tile_shape = tile_shape
-        ny, nx = (
-            int(math.ceil(float(N) / n)) for N, n in zip(box.shape, tile_shape.shape)
-        )
-        self._shape = shape_((ny, nx))
+        self._tiles = RoiTiles(box.shape, tile_shape)
         self._cache: Dict[Index2d, GeoBox] = {}
 
     @property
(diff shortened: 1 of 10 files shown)


odc/geo/geobox.py
```diff
@@ -237,7 +238,7 @@ def __and__(self, other) -> "GeoBox":
 
     def is_empty(self) -> bool:
         """Check if geobox is "empty"."""
-        return self._shape.shape == (0, 0)
+        return 0 in self._shape
```
gadomski, line 241: Just a stray thought -- would it help to make a `class EmptyGeoBox(GeoBox)` to make the empty case more explicit? Might not be worth the hassle but just curious what you think.

odc/geo/ui.py
```diff
@@ -73,7 +76,11 @@ def svg_base_map(
     # in pixels
     min_span = min(40, sz)
 
-    if span_x > span_y:
+    if min(span_x, span_y) < 1e-4:
```
gadomski: Could you add a code comment explaining why `1e-4` is a fine constant? I _think_ that its because we're assumed to be in web mercator here, but I'm not sure.

odc/geo/geobox.py
Kirill888: I think it will complicate type information without any real gain, in a statically typed language that might make sense, but in Python only people with mypy might "benefit" as you can type hint as "not empty geobox", but...

odc/geo/ui.py
Kirill888: kinda combination of numeric precision, round number and "what is likely empty anyway", this is roughly 11m at equator, so 1 sentinel pixel or there about, even if we had high-res map this would be way too zoomed in to make sense, maybe we should pick a bigger number.

now that I re-read it, it's not quite right, I don't think this will display something like `1x10_000` slice of sentinel correctly. i think this was meant to be `max` not `min`. I'll change it and add comment about what 1e-4 implies here.

Kirill888: reworked that and change tolerance check to be purely numeric based

gadomski (approved):

PR #37, opened 2022-04-21
TYPE: Add py.typed

### LINKED ISSUE #36: TYPE: missing py.typed
From mypy:
```
error: Skipping analyzing "odc.geo": module is installed, but missing library stubs or py.typed marker
```

snowman2:
Closes #36

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,6 +23,7 @@ classifiers =
     Programming Language :: Python :: 3.9
     Topic :: Software Development :: Libraries :: Python Modules
     Topic :: Scientific/Engineering :: GIS
+    Typing :: Typed
 
 [options]
 include_package_data = True
@@ -56,6 +57,8 @@ include =
   odc*
 
 [options.package_data]
+odc.geo =
+    py.typed
 odc.geo.data =
    *.geojson.xz
    *.css


setup.cfg
```diff
@@ -23,6 +23,7 @@ classifiers =
     Programming Language :: Python :: 3.9
     Topic :: Software Development :: Libraries :: Python Modules
     Topic :: Scientific/Engineering :: GIS
+    Typing :: Typed
```
Kirill888, line 26: So this tells mypy that types are present?

setup.cfg
```diff
@@ -56,6 +57,8 @@ include =
   odc*
 
 [options.package_data]
+odc.geo =
```
Kirill888, line 60: And this to include any stubs?

~1. we currently don't have stubs, should they be generated at packaging?~
2. should this also go into MANIFEST.in for source distribution?

Kirill888: Ok, I see, it's just an empty file marker.

snowman2: > should this also go into MANIFEST.in for source distribution?

I haven't needed to. The package data part seems to do the trick.

setup.cfg
snowman2: https://mypy.readthedocs.io/en/stable/installed_packages.html

> If you would like to publish a library package to a package repository yourself (e.g. on PyPI) for either internal or external use in type checking, packages that supply type information via type comments or annotations in the code should put a py.typed file in their package directory.

PR #41, opened 2022-04-27
BUG: Handle input CRS without EPSG

snowman2:
Without fix:
```
CRSError: Invalid projection: EPSG:None: (Internal Proj Error: proj_create: crs not found)
```


Kirill888 (approved):
thanks, clearly we need better test coverage for those :(

PR #42, opened 2022-04-27
REF: simplify CRS initialization logic

snowman2:
Minor tweak

### DIFF
--- a/odc/geo/crs.py
+++ b/odc/geo/crs.py
@@ -69,13 +69,11 @@ class CRS:
         :raises: :py:class:`pyproj.exceptions.CRSError`
         """
 
-        if isinstance(crs_spec, str):
-            self._crs, self._str, self._epsg = _make_crs(crs_spec)
-        elif isinstance(crs_spec, CRS):
-            self._crs = crs_spec._crs
-            self._epsg = crs_spec._epsg
-            self._str = crs_spec._str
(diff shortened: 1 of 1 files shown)


odc/geo/crs.py
```diff
@@ -69,13 +69,11 @@ def __init__(self, crs_spec: Any):
...
-        elif isinstance(crs_spec, CRS):
-            self._crs = crs_spec._crs
-            self._epsg = crs_spec._epsg
-            self._str = crs_spec._str
-        elif isinstance(crs_spec, _CRS):
+        if isinstance(crs_spec, CRS):
+            self._crs: _CRS = crs_spec._crs
+            self._epsg: Optional[int] = crs_spec._epsg
+            self._str: str = crs_spec._str
+        elif isinstance(crs_spec, (str, _CRS)):
```
Kirill888, line 76: Not too sure about this tweak to be honest. Original code dispatches on input type using most-likely first order. Most common is `str` in my opinion, doubt it makes much difference performance-wise, but it also makes it clear what types are supported and it what order type narrowing happens, yes that leads to some code duplication, but it's minimal..

I wouldn't mind accepting an `int` as a stand-in for `epsg` code maybe?

snowman2: That's fine with me. I will close this PR as the change doesn't really change much and you have a reason for why it is the way it is.

PR #45, opened 2022-05-12
Add GeoBox model docs and extra feature to xr_zeros

Kirill888:
- Page describing GeoBox
- Support dask arrays in xr_zeros
- Adding dask to test env
- Bumping version to 0.1.3


Kirill888:
@gadomski I'm going to merge this as I need this for `odc-stac` work, but please check docs in particular, we can patch those after release without issues.

Issue #51, opened 2022-06-08
Dask  colorize vmin vmax error


RichardScottOZ:
this failed

```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/tmp/ipykernel_19782/44836221.py in <cell line: 1>()
----> 1 newcheckdask = odc.geo.xr.colorize(newcheck['test'], cmap=newcmp,vmin=1,vmax=240)

~/exploracornzarr/lib/python3.8/site-packages/odc/geo/_rgba.py in colorize(x, cmap, attrs, clip, vmin, vmax)
    240 
    241         assert x.chunks is not None
--> 242         data = da.map_blocks(
    243             _impl,
    244             x.data,

~/exploracornzarr/lib/python3.8/site-packages/dask/array/core.py in map_blocks(func, name, token, dtype, chunks, drop_axis, new_axis, meta, *args, **kwargs)
    732         adjust_chunks = None
    733 
--> 734     out = blockwise(
    735         func,
    736         out_ind,

~/exploracornzarr/lib/python3.8/site-packages/dask/array/blockwise.py in blockwise(func, out_ind, name, token, dtype, adjust_chunks, new_axes, align_arrays, concatenate, meta, *args, **kwargs)
    274         from .utils import compute_meta
    275 
--> 276         meta = compute_meta(func, dtype, *args[::2], **kwargs)
    277     return new_da_object(graph, out, chunks, meta=meta, dtype=dtype)
    278 

~/exploracornzarr/lib/python3.8/site-packages/dask/array/utils.py in compute_meta(func, _dtype, *args, **kwargs)
    158                 return None
    159 
--> 160         if _dtype and getattr(meta, "dtype", None) != _dtype:
    161             with contextlib.suppress(AttributeError):
    162                 meta = meta.astype(_dtype)

~/exploracornzarr/lib/python3.8/site-packages/dask/delayed.py in __bool__(self)
    588 
    589     def __bool__(self):
--> 590         raise TypeError("Truth of Delayed objects is not supported")
    591 
    592     __nonzero__ = __bool__

TypeError: Truth of Delayed objects is not supported
```

this works - but as your docs suggest, gives strange results

```python
newcheckdask = odc.geo.xr.colorize(newcheck['test'], cmap=newcmp,vmin=1,vmax=240)
```

Have I done something wrong, or is there a bug?

Kirill888:
what's inside `newcmp`? Can you try with just a name of the colormap?

[...]

PR #54, opened 2022-07-05
Mostly various cleanups and more tests + GCPs helpers

### LINKED ISSUE #52: Implement robust cmap plotting
`xarray`'s plotting methods include a `robust=True` flag to set vmin/vmax to the 2nd and 98th percentile, respectively `ds.plot.imshow(robust=True)`. This ensures the colouramp doesn't overly account for outliers in the data.  Implementing the same feature in `ds.odc.add_to(robust=True)` would be very handy.

i.e. The equivalent of this:
```
vmin, vmax = np.nanpercentile(xx.band.data, [2, 98])
xx.band.odc.add_to(vmin=vmin, vmax=vmax)
```

Kirill888:
- Closes #52 
- maybe addresses #51
- Add `GeoBox.compat` to convert to datacube version of `GeoBox`
- In reproject use `nan` as fill value for `float32` outputs if output `nodata` is not configured
   - fixes boundary effects when plotting float32 input on a map
- In `colorize` handle `nodata` values properly for non float inputs
   - fixes boundary effects when plotting non-float inputs on a map
- In `colorize` drop `nodata` attribute on the output array
- In geobox display support non-epsg authority codes like `ESRI`
- Better Dask support for GeoBox, CRS and GeoBoxTiles classes (fixes hashing)
- Add GCP extraction utilities
- Add `Geometry[Point] -> XY[float]` conversion
- Deprecated `Geometry.__iter__` interface to follow `shapely` deprecation of that
- Fixes `.odc.add_to` for maps with custom projections


Kirill888:
@RichardScottOZ would you be able to check if this fixes your `colorize` workflow with custom colormap and Dask?

RichardScottOZ:
Kirill, will have a look in the next few days.

PR #62, opened 2022-08-21
Add GCPGeoBox

Kirill888:
Generalizes concept of `GeoBox` to include any bounded pixel plane with some well-defined mapping between pixel and world planes. Original `GeoBox` is a version of that with linear mapping `pix<->wld`.

This PR moves common properties of "geoboxes" into `GeoBoxBase` and adds `GCPGeoBox` class that handles ground control point based geoboxes.

- `xx.odc.geobox` will now return `GCPGeoBox` for GCP sources loaded with rioxarray
- `xr_coords|xr_zeros` can create xarrays from `GCPGeoBox`
- `xr_reproject` accepts GCP registered arrays as sources, but not as destinations as this is not exposed in rasterio
- `xx.odc.add_to(map)` also works for GCP based sources
- overlap computation logic has been updated to work either geobox type
- moved GeoBox html_repr logic into a separate class and file to enable sharing with GCPGeoBox

Other coincidental changes

- `Geometry` can now be constructed from geojson
- Adding more tests
- Fix `.dropna` for polygons with holes

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -31,6 +31,7 @@ Interfacing with :py:class:`xarray.DataArray` and :py:class:`xarray.Dataset` cla
    ODCExtensionDa.compress
    ODCExtensionDa.ydim
    ODCExtensionDa.xdim
+   ODCExtensionDa.nodata
 
    ODCExtensionDs
    ODCExtensionDs.to_rgba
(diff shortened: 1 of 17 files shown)


uchchwhash (approved):
Long PR, just skimmed through it, but looks good to me! 👍🏽

SpacemanPaul (approved):
It's a big PR and not an area I know much about, but GeoBox support for GCP based rasters sounds like a *very* useful enhancement.

PR #65, opened 2022-09-05
Add CRS.utm and support crs="utm" inputs in to_crs methods

### LINKED ISSUE #31: API: extend geobox slicing to support more input types
I think syntax like this should be possible and makes sense:

```python
geobox: GeoBox  = some_geobox(...)
poly: Geometry = some_geom(...)
assert geobox.crs == poly.crs

sub_geobox = geobox[poly]
assert geobox[poly] == geobox[poly.boundingbox]
```

If #25 is done then we can remove same crs constraint in which case

```python
assert geobox[poly] == geobox[poly.to_crs(geobox.crs).boundingbox]
```

Alternatively or in addition this could also be thought of as `intersection` operation and have the following syntax:

```python
sub_geobox: GeoBox = geobox & poly
sub_geobox: GeoBox = geobox & poly.boundingbox
```

Returned `sub_geobox` will enclose parts of the supplied geometry that overlap with the original `geobox` in a tight fashion

```python
assert sub_geobox.extent.contains(poly & geobox.extent)
```

### LINKED ISSUE #59: Add a `snap_to` method to GeoBoxes to enable coercing GeoBoxes to the same pixel alignment
I have two low-resolution GeoBoxes created by "zooming out" of two existing higher resolution dataset GeoBoxes:
```
lowres_geobox1 = ds1.odc.geobox.zoom_out(500)
lowres_geobox2 = ds2.odc.geobox.zoom_out(500)
```

Although these GeoBoxes share the same resolution and CRS, they do not share the same pixel alignment, resulting in pixels that do not align correctly:
image

I would like to be able to "snap" `lowres_geobox2` to the pixel alignment of `lowres_geobox1`, so that both of my low resolution grids align perfectly:
image

Snapping is currently supported when creating a new GeoBox directly (e.g. via `GeoBox.from_bbox`), however it would be useful to have a GeoBox method that would allow a user to perform snapping on existing GeoBoxes as well.

Kirill888:
## UTM convenience methods

1. `CRS.utm(some_geom)` picks appropriate utm zone for a given geometry and returns corresponding CRS object
2. When transforming geometry or geobox accept `crs="utm|utm-n|utm-s"` as a valid destination crs, meaning pick "appropriate" utm CRS for an object being transformed.
3. ~"appropriate UTM zone" is currently defined by the `lon,lat` of the centroid of the geometry~
4. UTM zones that overlap with `some_geom` are located using `pyproj`, then zone with biggest overlap is chosen

## Snap to

Adds `GeoBox.snap_to(other)`, this method adjust source geobox slightly by shifting footprint by a sub-pixel amount such that pixel edges align exactly with pixel edges of the `other` geobox. Method returns adjusted copy of self. Method only works when it is possible to align pixels by pure translation only - same projection, same resolution, same orientation. `ValueError` is raised when it's not possible.

Closes #59

## Enclosing

Related to `snap_to` somewhat. GeoBox construction method that creates a new "pixel grid compatible" geobox that minimally covers some area.

## Generalized Slicing into GeoBox

Overload `geobox[roi]` to support geometry types on input, footprint in the world can be converted to pixel space, from which pixel slice is computed.

Closes #31

### DIFF
--- a/odc/geo/crs.py
+++ b/odc/geo/crs.py
@@ -357,25 +357,35 @@ class CRSMismatchError(ValueError):
 @overload
 def norm_crs(crs: SomeCRS) -> CRS: ...
 @overload
+def norm_crs(crs: SomeCRS, ctx: Any) -> CRS: ...
+@overload
 def norm_crs(crs: None) -> None: ...
+@overload
+def norm_crs(crs: None, ctx: Any) -> None: ...
 # fmt: on
 
 
-def norm_crs(crs: MaybeCRS) -> Optional[CRS]:
+def norm_crs(crs: MaybeCRS, ctx=None) -> Optional[CRS]:
     """Normalise CRS representation."""
     if isinstance(crs, CRS):
         return crs
     if crs is None:
         return None
+    if isinstance(crs, str) and crs.lower() == "utm":
+        assert ctx is not None
+        return CRS.utm(ctx)
     return CRS(crs)
 
 
-def norm_crs_or_error(crs: MaybeCRS) -> CRS:
+def norm_crs_or_error(crs: MaybeCRS, ctx=None) -> CRS:
     """Normalise CRS representation, raise error if input is ``None``."""
     if isinstance(crs, CRS):
         return crs
     if crs is None:
         raise ValueError("Expect valid CRS")
+    if isinstance(crs, str) and crs.lower() == "utm":
+        assert ctx is not None
+        return CRS.utm(ctx)
     return CRS(crs)
 
 
--- a/odc/geo/crs.py
+++ b/odc/geo/crs.py
@@ -396,9 +396,24 @@ def norm_crs(crs: MaybeCRS, ctx=None) -> Optional[CRS]:
         return crs
     if crs is None:
         return None
-    if isinstance(crs, str) and crs.lower() == "utm":
-        assert ctx is not None
-        return CRS.utm(ctx)
+    if isinstance(crs, str):
+        _txt = crs.lower()
+        if _txt.startswith("utm"):
+            assert ctx is not None
+
+            utm_crs = CRS.utm(ctx)
+            if _txt == "utm":
+                return utm_crs
+
+            utm_zone = utm_crs.proj.utm_zone
+            epsg = utm_crs.epsg
+            assert utm_zone is not None
+            assert epsg is not None
+            if _txt == "utm-n" and utm_zone.endswith("S"):
+                utm_crs = CRS(epsg - 100)
+            elif _txt == "utm-s" and utm_zone.endswith("N"):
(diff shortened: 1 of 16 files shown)


odc/geo/crs.py
```diff
@@ -309,6 +311,38 @@ def result(x, y):
     def __dask_tokenize__(self):
         return ("odc.geo.crs.CRS", str(self))
 
+    @staticmethod
+    def utm(
```
snowman2, line 325: I believe this method handles scenarios when the UTM zone is irregular:

https://pyproj4.github.io/pyproj/stable/examples.html#find-utm-crs-by-latitude-and-longitude

It is used by geopandas and rioxarray in their `estimate_utm_crs` methods.

snowman2: See: https://github.com/geopandas/geopandas/blob/1821b6d72791ce1f7da32db5b22e8803a3dc44b9/geopandas/array.py#L770

Kirill888: great, thanks for that, I'll change the logic to use pyproj query

robbibt:
Hey @Kirill888, this looks fantastic - the UTM zone thing in particular looks really useful. One issue/question that occured to me however is that Landsat data is always packaged with UTM "North" zones (e.g. [EPSG:32656](https://epsg.io/32656)) even if the data itself lies in the "South" zone (e.g. [EPSG:32756](https://epsg.io/32756)) - if this was integrated into ODC-core and I tried native loading southern hemisphere data by supplying `crs="utm"`, would I risk having my data reprojected accidently? If so, I wonder if having an additional "utm-n" option could be useful, to force it to return the northern UTM zone even if the query is in the southern hemisphere (or maybe that's something that should be dealt with outside of `odc-geo`)

Kirill888:
@robbibt pretty sure it would perform reproject when going from north utm to south utm.

So maybe `utm-n`, `utm-s` should be an option. Although I'm switching this code to use pyproj query functionality so I'll need to investigate best way to construct "same zone and datum UTM CRS, but in north hemisphere", maybe `pyproj` already exposes that. 

But what you really need for `.load` is `crs="native"` and `crs="auto"` where `native` raises an error when there are more candidates than 1 and `auto` just picks most common.

snowman2:
Side note: It may also be helpful to look for these projections as well:
- https://epsg.org/crs_32600/WGS-84-UTM-grid-system-northern-hemisphere.html
- https://epsg.org/crs_32700/WGS-84-UTM-grid-system-southern-hemisphere.html

Kirill888:
> Side note: It may also be helpful to look for these projections as well:

hm, should we maybe use those when supplied region spans more than 2 UTM zones in the east-west direction?

ping: @robbibt, @SpacemanPaul

PR #70, opened 2022-10-14
PERF: CRS - use WKT for cache and lazy load EPSG code

snowman2:
Port of https://github.com/opendatacube/datacube-core/pull/1322

### DIFF
--- a/odc/geo/crs.py
+++ b/odc/geo/crs.py
@@ -9,6 +9,7 @@ from typing import (
     Callable,
     Dict,
     List,
+    Literal,
     Optional,
     Tuple,
     Union,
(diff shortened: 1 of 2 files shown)


Kirill888:
@snowman2 thanks for this, but I'm not too sure this is the right way to approach this problem. I think `rasterio.crs.CRS` is a common enough input type and we should handle it in the `_make_crs` cached path, rather than handling it in a "generic" section. `rasterio.crs.CRS` is a hashable object so can be used as a key directly, only issue is that I want `odc-geo` to be independent from `rasterio`, so it might be a bit trickier to implement because of that.

I'm guessing that a performant way to handle `rasterio.crs.CRS` object would be to query `.is_epsg_code` property before attempting to call `.to_epsg`, also with those things going through cache performance impact will be reduced.

PR #72, opened 2022-10-24
Delay epsg determination as much as possible

Kirill888:
Introduce special state _epsg=0, that implies "epsg code is not known yet", `_epsg=None` means "epsg code is known to be absent".

Force all CRS construction code paths through cache.

### DIFF
--- a/odc/geo/crs.py
+++ b/odc/geo/crs.py
@@ -121,9 +121,7 @@ class CRS:
         elif hasattr(crs_spec, "to_wkt"):
             self._crs, self._str, self._epsg = _make_crs(crs_spec)
         else:
-            raise CRSError(
-                "Expect string or any object with `.to_epsg()` or `.to_wkt()` methods"
-            )
+            raise CRSError(f"Unexpected input encountered: {crs_spec}")
 
     def __getstate__(self):
         return {"crs_str": self._str}
@@ -152,7 +150,7 @@ class CRS:
         """
         EPSG Code of the CRS or ``None``.
         """
-        if self._epsg == 0:
+        if self._epsg == EPSG_UNSET:
             self._epsg = self._crs.to_epsg()
         return self._epsg
 
@@ -228,7 +226,7 @@ class CRS:
 
         :returns: ``("", "")`` when not available
         """
-        if self._epsg is not None and self._epsg > 0:
+        if self._epsg:
             return ("EPSG", self._epsg)
 
         if (r := self._crs.to_authority()) is not None:
@@ -259,8 +257,11 @@ class CRS:
         if self._crs is other._crs:
             return True
 
-        if self.epsg is not None and other.epsg is not None:
-            return self.epsg == other.epsg
+        if self._epsg and other._epsg:
+            return self._epsg == other._epsg
+
+        if self._str == other._str:
+            return True
 
         return self._crs == other._crs
 
--- a/odc/geo/crs.py
+++ b/odc/geo/crs.py
@@ -2,14 +2,17 @@
 #
 # Copyright (c) 2015-2020 ODC Contributors
 # SPDX-License-Identifier: Apache-2.0
+import collections
 import warnings
 from typing import (
     TYPE_CHECKING,
     Any,
     Callable,
     Dict,
+    Hashable,
     List,
     Optional,
+    Protocol,
     Tuple,
     Union,
     overload,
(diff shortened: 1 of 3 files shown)


Kirill888:
@snowman2 how about this change? Do you think that will fix performance issues you have observed with non-epsg SRSs?

odc/geo/crs.py
```diff
@@ -93,22 +118,9 @@ def __init__(self, crs_spec: Any):
...
-            try:
-                wkt = crs_spec.to_wkt()
-            except AttributeError:
-                wkt = None
-            if wkt is not None:
-                self._crs, self._str, self._epsg = _make_crs(wkt)
-                return
-
             raise CRSError(
                 "Expect string or any object with `.to_epsg()` or `.to_wkt()` methods"
```
snowman2:
```suggestion
               f"Unexpected input encountered: {crs_spec}"
```

odc/geo/crs.py
```diff
@@ -26,32 +29,54 @@
 
 SomeCRS = Union[str, int, "CRS", _CRS, Dict[str, Any]]
 MaybeCRS = Union[SomeCRS, Unset, None]
+EPSG_UNSET = 0
```
snowman2, line 32: To be safe, I am thinking `-1` may be a good choice.
```suggestion
EPSG_UNSET = -1
```

odc/geo/crs.py
```diff
@@ -140,14 +152,16 @@ def to_epsg(self) -> Optional[int]:
         """
         EPSG Code of the CRS or ``None``.
         """
+        if self._epsg == 0:
```
snowman2:
```suggestion
        if self._epsg == EPSG_UNSET:
```

snowman2:
Part of the reason `is False` was used in #70 was because it skips the equality check as it is not necessary. Using `==` instead of `is` invokes the `__eq__` method of the CRS class which attempts to create a CRS. This can cause a performance hit as PROJ has to go through a lot of logic before determining that it isn't a valid CRS.

snowman2:
```python
In [1]: from pyproj import CRS

In [2]: cc = CRS(4326)

In [4]: %%timeit
   ...: 
   ...: cc == -1
   ...: 
   ...: 
7.54 ms ± 252 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [5]: %%timeit
   ...: 
   ...: cc is False
   ...: 
   ...: 
59.1 ns ± 5.09 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

```

snowman2:
> Do you think that will fix performance issues you have observed with non-epsg SRSs?

Yes, I believe it will help. Thanks :+1:

Kirill888:
> Part of the reason `is False` was used in #70 was because it skips the equality check as it is not necessary. Using `==` instead of `is` invokes the `__eq__` method of the CRS class which attempts to create a CRS. This can cause a performance hit as PROJ has to go through a lot of logic before determining that it isn't a valid CRS.

My main issue with using `False|None|<int>` is that it complicates type a bit, and it's not really obvious which is which between `None` and `False`. The `CRS.__eq__` should not trigger with the current implementation as comparison is done with `None|int` type

snowman2:
I gave it a try:

```python
In [1]: from odc.geo import CRS

In [2]: cc = CRS(4326)

In [3]: %%timeit
   ...: 
   ...: cc == -1
   ...: 
   ...: 
5.07 ms ± 202 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [4]: %%timeit
   ...: 
   ...: cc is False
   ...: 
   ...: 
31.4 ns ± 3.22 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

snowman2:
Looks like it attempts to construct a CRS if it is not a CRS each time: https://github.com/opendatacube/odc-geo/blob/85c90a232d0888408dd1e53e1a913239d6480cf9/odc/geo/crs.py#L238-L243

snowman2:
Actually, need to update to use this PR ...

snowman2:
You are correct, this seems okay:
```python
In [5]: %%timeit
   ...: 
   ...: cc.to_epsg()
   ...: 
   ...: 
141 ns ± 1.87 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

snowman2:
And this looks good too:
```python
In [7]: %%timeit
   ...: 
   ...: CRS('PROJCS["unnamed",GEOGCS["Unknown datum based upon the custom spher
   ...: oid",DATUM["Not specified (based on custom spheroid)",SPHEROID["Custom 
   ...: spheroid",6371007.181,0]],PRIMEM["Greenwich",0],UNIT["degree",0.0174532
   ...: 925199433]],PROJECTION["Sinusoidal"],PARAMETER["longitude_of_center",0]
   ...: ,PARAMETER["false_easting",0],PARAMETER["false_northing",0],UNIT["Meter
   ...: ",1],AXIS["Easting",EAST],AXIS["Northing",NORTH]]')
   ...: 
   ...: 
1.06 µs ± 19.9 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

odc/geo/crs.py
Kirill888: I'm keeping it at zero for ease of `if self._epsg:` in `__eq__` and `.authority` functions, basically if `self._epsg` is `falsy` then it's either unknown or known to be absent.

snowman2: Okay, that is likely fine.

Kirill888:
> Looks like it attempts to construct a CRS if it is not a CRS each time:
> 
> https://github.com/opendatacube/odc-geo/blob/85c90a232d0888408dd1e53e1a913239d6480cf9/odc/geo/crs.py#L238-L243

I guess we could avoid that when `rhs` is `int`, and maybe avoid it when `rhs` is string.

Kirill888:
Although, with every input into CRS constructor going through the cache path, it's less of an issue.

snowman2 (approved):
Thanks 👍

PR #74, opened 2022-10-25
BUG: Use typing.Hashable instead of collections.Hashable

### LINKED ISSUE #73: 0.3.1 broke compatibility with Python 3.10
This is somewhat confusing, but I think Python 3.10 deprecated classes that are generic within collections (Collections Abstract Base Classes): https://docs.python.org/3/whatsnew/3.10.html#collections-abc. 

That means 2b32ed6bca5949dcee313f5b965ee15e5a7d0e88 and your latest release results in the following error when using Python 3.10:

```
...
File "/usr/local/lib/python3.10/site-packages/odc/geo/crs.py", line 53, in _make_crs_key
    if isinstance(crs_spec, collections.Hashable):
AttributeError: module 'collections' has no attribute 'Hashable'
```

snowman2:
Closes #73 

https://docs.python.org/3/library/typing.html#typing.Hashable

> An alias to [collections.abc.Hashable](https://docs.python.org/3/library/collections.abc.html#collections.abc.Hashable).

### DIFF
--- a/odc/geo/crs.py
+++ b/odc/geo/crs.py
@@ -2,7 +2,6 @@
 #
 # Copyright (c) 2015-2020 ODC Contributors
 # SPDX-License-Identifier: Apache-2.0
-import collections
 import warnings
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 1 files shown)


Kirill888 (approved):
thanks for that

Issue #102, opened 2023-06-02
Make `.odc.add_to()` use input array's `.name` attribute if available 


robbibt:
The `.odc.add_to()` method allows users to specify a `name` that is used to name layers in Folium and Ipyleaflet Layer Controls. If `name` is left to its default `None`, then we get either an awkard auto-generated layer name (Folium):
![image](https://github.com/opendatacube/odc-geo/assets/17680388/85522072-1e9a-4a24-9c11-b0985ab6dbbe)

Or no layer name (Ipyleaflet):
![image](https://github.com/opendatacube/odc-geo/assets/17680388/8f6f4093-5dc3-4418-826d-7e35d8e0bbc7)

Xarray datasets can however commonly include a `.name` attribute, for example here:
![image](https://github.com/opendatacube/odc-geo/assets/17680388/f510ebb3-d47a-42ff-83cb-38a6d094b27b)

It would be great for the default of `name=None` to use this `.name` attribute by default if it exists.

PR #103, opened 2023-06-02
Use xarray `.name` attribute as Folium/Ipyleaflet layer name if it exists

robbibt:
This PR addresses #102 by updating the `odc.add_to()` method to use the `.name` attribute of the input array if it exists AND `name=None`.


Kirill888 (approved):
👍

PR #105, opened 2023-06-07
Updated test requirements

SpacemanPaul:
Tests do not pass after running the following in the repository in a clean venv:

```
pip install -e .[tests]
pytests
```

mock, dask and geopandas are need for running tests so added to test requirements in `setup.cfg`.


snowman2:
Thoughts about using [unittest.mock](https://docs.python.org/3/library/unittest.mock.html) instead of mock?

SpacemanPaul:
> Thoughts about using [unittest.mock](https://docs.python.org/3/library/unittest.mock.html) instead of mock?

Personally, I have no strong preference either way - as long as I can pip install odc-geo[test] and successfully run the tests.

Kirill888:
@snowman2 thanks for the suggestion, looks like we already used it in some of the test files but not in others.

Kirill888 (approved):

PR #114, opened 2023-12-17
Add new `odc.mask` and `odc.crop` methods for masking and cropping `xarray` data by geometries

robbibt:
This PR introduces a new `.odc.crop` method that supports:

- Clipping an `xarray.Dataset` or `xr.DataArray` to the spatial extent of a datacube `Geometry` polygon
- Optionally apply the polygon as a mask by rasterizing it to the new geobox

image

This will be very useful for pixel drill workflows that regularly involve a sequence of loading data, cropping to a certain extent using `.sel()`, then rasterizing a polygon and applying it as a mask.

I've tested that this works on both datasets and data arrays, and have included a check that the polygon overlaps with the geobox extent. Would welcome any other feedback to make this more robust!

### DIFF
--- a/odc/geo/_xr_interop.py
+++ b/odc/geo/_xr_interop.py
@@ -34,7 +34,9 @@ from .geobox import Coordinate, GeoBox
 from .geom import Geometry
 from .math import affine_from_axis, maybe_int, resolution_from_affine
 from .overlap import compute_output_geobox
+from .roi import roi_is_empty
 from .types import Resolution, xy_
+
 
 # pylint: disable=import-outside-toplevel
 # pylint: disable=too-many-lines
@@ -266,10 +268,6 @@ def mask(xx: XrT, poly: Geometry, all_touched: bool = True) -> XrT:
         A :py:class:`~xarray.Dataset` or :py:class:`~xarray.DataArray`
         masked by ``poly``.
     """
-
-    # Reproject `poly` to match `xx`
-    poly = poly.to_crs(crs=xx.odc.crs)
-
     # Rasterise `poly` into geobox of `xx`
     rasterized = rasterize(
         poly=poly,
@@ -279,6 +277,13 @@ def mask(xx: XrT, poly: Geometry, all_touched: bool = True) -> XrT:
 
     # Mask data outside rasterized `poly`
     xx_masked = xx.where(rasterized.data)
+
+    # Remove nodata attribute from arrays
+    if isinstance(xx_masked, xarray.Dataset):
+        for var in xx_masked.data_vars:
+            xx_masked[var].attrs.pop("nodata", None)
+    else:
+        xx_masked.attrs.pop("nodata", None)
 
     return xx_masked
 
@@ -309,29 +314,28 @@ def crop(
         A :py:class:`~xarray.Dataset` or :py:class:`~xarray.DataArray`
         cropped and optionally masked to the spatial extent of ``poly``.
     """
-    # Reproject `poly` to match `xx`
-    poly = poly.to_crs(crs=xx.odc.crs)
+    # Create new geobox with pixel grid of `xx` but enclosing `poly`.
+    poly_geobox = xx.odc.geobox.enclosing(poly)
 
-    # Verify that `poly` overlaps with `xx` extent
-    if not poly.intersects(xx.odc.geobox.extent):
+    # Calculate ROI slices into `xx` for intersection between both geoboxes.
+    roi = xx.odc.geobox.overlap_roi(poly_geobox)
+
+    # Verify that `poly` overlaps with `xx` by checking if the returned
+    # ROI is empty
+    if roi_is_empty(roi):
         raise ValueError(
             "The supplied `poly` must overlap spatially with the extent of `xx`."
         )
 
-    # Optionally mask data outside rasterized `poly`
-    if apply_mask:
-        xx = mask(xx, poly, all_touched=all_touched)
-
-    # Identify spatial dims and raise error if None
+    # Crop spatial dims of `xx` using ROI
     sdims = spatial_dims(xx)
     if sdims is None:
         raise ValueError("Can't locate spatial dimensions")
-
-    # Crop `xx` to the bounding box of `poly`. First create new geobox
-    # with the same pixel grid as `xx` but enclosing `poly`. Then
-    # calculate slices into `xx` for the intersection between both geoboxes.
-    roi = xx.odc.geobox.overlap_roi(xx.odc.geobox.enclosing(poly))
     xx_cropped = xx.isel({sdims[0]: roi[0], sdims[1]: roi[1]})
+
+    # Optionally mask data outside rasterized `poly`
+    if apply_mask:
+        xx_cropped = mask(xx_cropped, poly, all_touched=all_touched)
 
     return xx_cropped
 
--- a/odc/geo/_xr_interop.py
+++ b/odc/geo/_xr_interop.py
@@ -248,7 +248,9 @@ def assign_crs(
     return xx
 
 
-def mask(xx: XrT, poly: Geometry, all_touched: bool = True) -> XrT:
+def mask(
+    xx: XrT, poly: Geometry, invert: bool = False, all_touched: bool = True
+) -> XrT:
     """
     Apply a polygon geometry as a mask, setting all xr.Dataset
     or xr.DataArray pixels outside the rasterized polygon to ``NaN``.
@@ -257,22 +259,30 @@ def mask(xx: XrT, poly: Geometry, all_touched: bool = True) -> XrT:
        :py:class:`~xarray.Dataset` or :py:class:`~xarray.DataArray`.
 
     :param poly:
-       Geometry shape used to mask ``xx``.
+       A :py:class:`odc.geo.geom.Geometry` polygon used to mask ``xx``.
+
+    :param invert:
+        Whether to invert the mask before applying it to ``xx``. If
+        ``True``, only pixels inside of ``poly`` will be masked.
 
     :param all_touched:
-        If ``True``, all pixels touched by ``poly`` will remain unmasked.
-        If ``False``, only pixels whose center is within the polygon or
-        that are selected by Bresenham's line algorithm will remain unmasked.
+        If ``True``, the rasterize step will burn in all pixels touched
+        by ``poly``. If ``False``, only pixels whose centers are within
+        the polygon or that are selected by Bresenham's line algorithm
+        will be burned in.
 
     :return:
         A :py:class:`~xarray.Dataset` or :py:class:`~xarray.DataArray`
         masked by ``poly``.
+
+    ..seealso:: :py:meth:`odc.geo.xr.rasterize`
     """
     # Rasterise `poly` into geobox of `xx`
     rasterized = rasterize(
         poly=poly,
         how=xx.odc.geobox,
         all_touched=all_touched,
+        value_inside=not invert,
     )
 
     # Mask data outside rasterized `poly`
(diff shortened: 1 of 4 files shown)


Kirill888:
This is great, thanks Robbi.

Can you please add this method to docs, here:

https://github.com/opendatacube/odc-geo/blob/6ce7524b6808457418b2c8bcecf01c918ce6458a/docs/api.rst?plain=1#L39-L52

odc/geo/_xr_interop.py
```diff
@@ -245,6 +245,59 @@ def assign_crs(
...
+    poly = poly.to_crs(crs=xx.odc.crs)
+    bbox = poly.boundingbox
+
+    # Verify that `poly` overlaps with `xx` extent
+    if poly.intersects(xx.odc.geobox.extent):
+        # Crop `xx` to the bounding box of `poly`
+        ydim, xdim = spatial_dims(xx)
+        xx_cropped = xx.sel(
+            {xdim: slice(bbox.left, bbox.right), ydim: slice(bbox.top, bbox.bottom)}
+        )
```
Kirill888: I think this is best covered by `GeoBox.enclosing` followed by geobox intersect, this should give you an roi of the original image overlapping with the geometry, which you can then use with `.isel` to select a sub-region.

odc/geo/_xr_interop.py
```diff
@@ -768,6 +822,8 @@ def to_rgba(
     ) -> xarray.DataArray:
         return to_rgba(self._xx, bands=bands, vmin=vmin, vmax=vmax)
 
+    crop = _wrap_op(crop)
```
robbibt: I didn't know whether to add this to just the "ODC extension base class" or both the "ODCExtensionDa" and "ODCExtensionDs". I went with the second option, but can easily change to the first one

odc/geo/_xr_interop.py
Kirill888: Combination of these two should give you a slice into original image covering given geometry:

https://github.com/opendatacube/odc-geo/blob/6ce7524b6808457418b2c8bcecf01c918ce6458a/odc/geo/geobox.py#L620

https://github.com/opendatacube/odc-geo/blob/6ce7524b6808457418b2c8bcecf01c918ce6458a/odc/geo/geobox.py#L657

Kirill888: in fact we should overload `overlap_roi` to allow geometry on input.

odc/geo/_xr_interop.py
Kirill888: if it works on both, it should be only in base class

odc/geo/_xr_interop.py
```diff
@@ -245,6 +245,59 @@ def assign_crs(
...
+        cropped and optionally masked to the spatial extent of ``poly``.
+    """
+    # Reproject `poly` to match `xx` and calculate bounding box
+    poly = poly.to_crs(crs=xx.odc.crs)
+    bbox = poly.boundingbox
+
+    # Verify that `poly` overlaps with `xx` extent
+    if poly.intersects(xx.odc.geobox.extent):
+        # Crop `xx` to the bounding box of `poly`
+        ydim, xdim = spatial_dims(xx)
```
Kirill888: mypy complains here about not handling the case when spatial_dims returns None.

odc/geo/_xr_interop.py
```diff
@@ -245,6 +245,59 @@ def assign_crs(
     return xx
```
Kirill888, line 248: pylint complains about too-may-lines, add disable line for that at the top of the file
`#pylint: disable=<message-name>`

odc/geo/_xr_interop.py
robbibt: Something like this?
```
xx.odc.geobox.overlap_roi(xx.odc.geobox.enclosing(poly))
>>> (slice(118, 285, None), slice(132, 276, None))
```

The output tuple doesn't include coordinate names, and I get `ValueError: Unsupported key-type <class 'tuple'>` if I try `xx[roi]`. Do I still need to grab coord names with `spatial_dims` or is there any way to pass an ROI to `isel` directly?

odc/geo/_xr_interop.py
robbibt: What would you recommend for that case? Just raise an exception? I 'm trying to think of a scenario where we'd expect this functionality to work with undefined spatial dims...

odc/geo/_xr_interop.py
robbibt: This works, but would be nice to avoid the `spatial_dims` step:
```
roi = xx.odc.geobox.overlap_roi(xx.odc.geobox.enclosing(poly))
ydim, xdim = spatial_dims(xx)
xx_cropped = xx.isel({ydim: roi[0], xdim: roi[1]})
```

robbibt: Ah... the problem is that you can only use an ROI to index a `xr.DataArray`, not an `xr.Dataset`...

image

odc/geo/_xr_interop.py
Kirill888: assert is enough for mypy usualy, like

```python
sdims = spatial_dims(xx)
assert sdims
ydim, xdim = sdims
```

but here we should probably use `ydim` property that returns an integer index of `y` dimension, instead of dimension names.

odc/geo/_xr_interop.py
Kirill888: instead of dimension names there is `.odc.ydim` which tells you which tells you where `y,x` dims are, always next to each other, from that one can construct `roi` object that slices to all dimensions.

Kirill888:
I think it might be better to expose 2 methods:

* `.mask(geom, ...options) -> xr.DataArray | xr.Dataset` returning masked result in the source geobox
* `.crop(geom, ...options) -> xr.DataArray | xr.Dataset` returning cropped and possibly masked result

@robbibt also when happy with changes please rebase it all to a single commit:

1. `git reset --soft HEAD~<number of commits to merge into one>`
2. commit everything again as a single change this time
3. push to github: `git push --force-with-lease`

odc/geo/_xr_interop.py
Kirill888: This is probably common enough functionality that needs to be exposed on `.odc` accessor

* Input: 2d slice into spatial dimensions 
* Output:
   * Nd slice into data variable(s) with extra dimensions filled with `:` slices
   * New Dataset/DataArray sliced into geo dims

So `np.s_[100:200, :30] -> np.s_[:, 100:200, :30]` in the common case, but for rgb data it would be `-> np.s_[:, 100:200, :30, :]`.

robbibt:
> I think it might be better to expose 2 methods:
> 
> * `.mask(geom, ...options) -> xr.DataArray | xr.Dataset` returning masked result in the source geobox
> * `.crop(geom, ...options) -> xr.DataArray | xr.Dataset` returning cropped and possibly masked result
> 
> @robbibt also when happy with changes please rebase it all to a single commit:
> 
> 1. `git reset --soft HEAD~<number of commits to merge into one>`
> 2. commit everything again as a single change this time
> 3. push to github: `git push --force-with-lease`

I agree, both a `mask` and `crop` method would be useful. I'll have another shot soon and will rebase everything after.

odc/geo/_xr_interop.py
```diff
@@ -245,6 +246,96 @@ def assign_crs(
...
+        If ``False``, only pixels whose center is within the polygon or
+        that are selected by Bresenham's line algorithm will remain unmasked.
+
+    :return:
+        A :py:class:`~xarray.Dataset` or :py:class:`~xarray.DataArray`
+        masked by ``poly``.
+    """
+
+    # Reproject `poly` to match `xx`
+    poly = poly.to_crs(crs=xx.odc.crs)
```
Kirill888: This is done internally by `rasterize`, so it's best not to repeat it here. Change in coordinate system for a geometry is often a "dangerous" thing that can break geometry. While `rasterize` doesn't currently do anything different than this code, it's best to keep number of places that perform such conversion to a minimum.

What `rasterize` **should** do when given a geometry in a projection other than what's desired on output

1. Call `.to_crs` with appropriate resolution computed from mask resolution
2. Check if any nodes failed to map from source to destination CRS, and if so, try again with a more costly approach which is not currently implemented, a kind of `robust_and_costly_to_crs(geom, crs, clip_region)`.

odc/geo/_xr_interop.py
```diff
@@ -245,6 +246,96 @@ def assign_crs(
...
+
+    # Rasterise `poly` into geobox of `xx`
+    rasterized = rasterize(
+        poly=poly,
+        how=xx.odc.geobox,
+        all_touched=all_touched,
+    )
+
+    # Mask data outside rasterized `poly`
+    xx_masked = xx.where(rasterized.data)
```
Kirill888, line 289: So, `xx.where` in that form always changes output to `float{64|32}+nan` representation. Which is a valid and often desired output format, but not always. One might also expect to retain existing pixel data type with `nodata` on output.

We should certainly support both. But it's less clear to me which one should be a default

* Personally I would expect to get `int16, nodata=-9999` on output if that was given on input, but I can see how `float32, nan` and even `float64, nan` can be also desired.
* But I can also see that some would expect and prefer to see `float32, nan` output for `mask` operation. 

In the code as it stands we should at least ensure that `nodata` attribute is not present on `xx_masked` output array, I still don't have any intuition for `xarray` attributes, when they stay and when they go...

odc/geo/_xr_interop.py
```diff
@@ -245,6 +246,96 @@ def assign_crs(
...
+    # Reproject `poly` to match `xx`
+    poly = poly.to_crs(crs=xx.odc.crs)
+
+    # Verify that `poly` overlaps with `xx` extent
+    if not poly.intersects(xx.odc.geobox.extent):
+        raise ValueError(
+            "The supplied `poly` must overlap spatially with the extent of `xx`."
+        )
+
+    # Optionally mask data outside rasterized `poly`
```
Kirill888: Order of operations matter here for performance/RAM usage here.

Mask should be applied AFTER cropping, not before.

odc/geo/_xr_interop.py
```diff
@@ -245,6 +246,96 @@ def assign_crs(
...
+        cropped and optionally masked to the spatial extent of ``poly``.
+    """
+    # Reproject `poly` to match `xx`
+    poly = poly.to_crs(crs=xx.odc.crs)
+
+    # Verify that `poly` overlaps with `xx` extent
+    if not poly.intersects(xx.odc.geobox.extent):
+        raise ValueError(
+            "The supplied `poly` must overlap spatially with the extent of `xx`."
+        )
```
Kirill888: Same story as before: it's dangerous to perform coordinate transform, if it can be avoided by pushing into more basic operations like `.enclosing`, then it should be. You'd then check if overlap_roi returns a null set and raise error then.

odc/geo/_xr_interop.py
```diff
@@ -657,6 +748,13 @@ def geobox(self) -> Optional[SomeGeoBox]:
         """Query :py:class:`~odc.geo.geobox.GeoBox` or :py:class:`~odc.geo.gcp.GCPGeoBox`."""
         return self._state.geobox
 
+    @property
```
Kirill888, line 768: where did this come from? bad merge or something?

odc/geo/_xr_interop.py
robbibt: Cool, I didn't realise it was already handled in `rasterize`. Removed and everything appears to still work nicely.

odc/geo/_xr_interop.py
robbibt: Done

odc/geo/_xr_interop.py
robbibt: OK, have revised to remove the coordinate transform, and simply run `enclosing` > `overlap_roi` then verify the ROI isn't empty using `roi_is_empty`:

https://github.com/opendatacube/odc-geo/blob/crop/odc/geo/_xr_interop.py#L330-L341

odc/geo/_xr_interop.py
robbibt: Yeah, this occurred to me too - I definitely think both options should be supported. My take would be that "most" of our more general users simply want a way to quickly set pixels outside of a polygon to `NaN` - which is a very common workflow for things like pixel drills or zonal statistics. Getting back `int16, nodata=-9999` is definitely useful but perhaps for a smaller number of more advanced users/workflows. I would be in favour of the simpler `NaN` option being the default.

The approach taken in `dea_tools.datahandling.load_ard` seems like it could be a good template: https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/Tools/dea_tools/datahandling.py#L214-L221
- `dtype="auto"`: When 'auto' is used, the data will be converted to `float32` if masking is used, otherwise data will be returned in the native data type of the data.
- `dtype="native"`: Data will be returned in the native data type of the data, including native nodata value
- `dtype='float{16|32|64}'`: Custom dtype

I guess given `mask` always applies a mask, we probably don't need "auto"... so a default of "float32" with an option for "native" could work.

robbibt: For the moment, I've added some code to strip nodata attributes:
https://github.com/opendatacube/odc-geo/blob/crop/odc/geo/_xr_interop.py#L291-L296

odc/geo/_xr_interop.py
robbibt: Hmm, I think I accidently went one commit to far when running `git reset --soft HEAD~<number of commits to merge into one>`.  Although it confuses me that the same code is in the `crop` branch and `develop` but is still being detected as a diff...
https://github.com/opendatacube/odc-geo/blob/crop/odc/geo/_xr_interop.py#L768-L773
https://github.com/opendatacube/odc-geo/blob/develop/odc/geo/_xr_interop.py#L660-L665

Kirill888 (approved):
@robbibt thanks for this, I think we should merge this as is, we can do further interface polishing,  before next release.

robbibt:
> @robbibt thanks for this, I think we should merge this as is, we can do further interface polishing, before next release.

Thanks @Kirill888 - should I combine the commits into one again? And what merge option do you use here? 

image

Kirill888:
I usually go with Rebase and merge for cleaner history

PR #116, opened 2024-01-04
Add new `.odc.explore()` method for automatically plotting raster data on an interactive map 

### LINKED ISSUE #104: Add a `.odc.explore()` method for automatically plotting raster data on an interactive map
Geopandas recently added an extremely useful `.explore()` method that allows users to quickly plot and inspect data on an interactive map: https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.explore.html

Being able to quickly plot and inspect raster data on an interactive map would be incredibly useful, and would avoid a common workflow of having to export to GeoTIFF and load data into a GIS just to inspect it closely.

Luckily with `odc-geo` we can already do something like this:
```
import folium
import rioxarray
import odc.geo.xr 

ds = rioxarray.open_rasterio(
    "https://dea-public-data-dev.s3-ap-southeast-2.amazonaws.com/derivative/dea_intertidal/sample_data/160/2019-2021/DEV_160_2019_2021_elevation.tif",
    masked=True,
).squeeze("band")

# Create folium Map
m = folium.Map()

# Reproject to 3857
ds_reproj = ds.odc.reproject("epsg:3857")

# Add to map
ds_reproj.odc.add_to(m)

# Zoom map to bounds
m.fit_bounds(ds.odc.map_bounds())
display(m)
```

But this adds a lot of boilerplate code that makes this harder than it needs to be.

**Proposal**
Instead, we could add a new `.odc.explore()` method to `odc-geo` that mimics the `.explore()` functionality of Geopandas. In my mind, calling `ds.odc.explore()` would automatically:
- Create a Folium or Ipyleaflet map
- Reproject data to "epsg:3857"
- Add the reprojected data to the map 
- Fit the map bounds to the data
- Display the map

I'd be keen to have a go at adding something like this, but I'd love to know if there's any obvious issues with this proposal. One thing I can think of is that neither `folium` or `ipyleaflet` are currently included as a requirement for `odc-geo` (I imagine to keep it as small and manageable as possible), so this may be a blocker to adding a feature that depends on those packages?

robbibt:
This PR addresses #104 by adding a new `.explore()` method for automatically plotting raster data into an interactive map. This is designed to replicate the extremely useful `.explore()` method from Geopandas: https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.explore.html

Essentially `explore` combines `to_rgba` and `add_to` from this repo into a single easy to use method.

Supports single band `xr.DataArray` inputs:
image

Automatically converts `xr.Dataset` inputs to RGB:
image

And easily passing in custom mapping params like basemaps:
image

### DIFF
--- a/odc/geo/_compress.py
+++ b/odc/geo/_compress.py
@@ -19,6 +19,17 @@ _fmt_info = {
 }
 
 
+def _verify_can_compress(xx: xr.DataArray):
+    """Returns error if array dimensions are not suitable for compress"""
+    if xx.ndim > 2:
+        xx = xx.squeeze()
+
+    if xx.ndim not in (2, 3):
+        raise ValueError(
+            f"Expected a 2 or 3 dimensional array; got {xx.ndim} dimensions {xx.dims}."
+        )
+
+
 def _compress_image(im: np.ndarray, driver="PNG", **opts) -> bytes:
     if im.ndim > 2:
         im = np.squeeze(im)
@@ -26,11 +37,9 @@ def _compress_image(im: np.ndarray, driver="PNG", **opts) -> bytes:
     if im.ndim == 3:
         h, w, nc = im.shape
         bands = np.transpose(im, axes=(2, 0, 1))  # Y,X,B -> B,Y,X
-    elif im.ndim == 2:
+    else:
         (h, w), nc = im.shape, 1
         bands = im.reshape(nc, h, w)
-    else:
-        raise ValueError(f"Expect 2 or 3 dimensional array got: {im.ndim}")
 
     rio_opts = {
         "width": w,
@@ -76,6 +85,9 @@ def compress(
       Pixel value to use for transparent pixels, useful for jpeg output.
 
     """
+    # Raise error early if xx has unsuitable dims
+    _verify_can_compress(xx)
+
     fmt = "png"
     opts = {}
     if len(args) >= 1:
--- a/odc/geo/_compress.py
+++ b/odc/geo/_compress.py
@@ -37,9 +37,13 @@ def _compress_image(im: np.ndarray, driver="PNG", **opts) -> bytes:
     if im.ndim == 3:
         h, w, nc = im.shape
         bands = np.transpose(im, axes=(2, 0, 1))  # Y,X,B -> B,Y,X
-    else:
+    elif im.ndim == 2:
         (h, w), nc = im.shape, 1
         bands = im.reshape(nc, h, w)
+    else:
+        raise ValueError(
+            f"Expected a 2 or 3 dimensional array; got {im.ndim} dimensions."
+        )
 
     rio_opts = {
         "width": w,
--- a/odc/geo/_map.py
+++ b/odc/geo/_map.py
@@ -58,6 +58,7 @@ def add_to(
     name: Optional[str] = None,
     fmt: str = "png",
     max_size: int = 4096,
+    resampling: str = "nearest",
     # jpeg options:
     transparent_pixel: Optional[Tuple[int, int, int]] = None,
     # RGB conversion parameters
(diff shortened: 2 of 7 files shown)


robbibt:
@Kirill888 I wasn't sure whether to define the main `explore` func in `_map.py` or `_xr_interop.py`. It might be better in `_xr_interop.py` so we can more easily include it in the docs, e.g. `odc.geo.xr.explore`?

odc/geo/_map.py
```diff
@@ -153,3 +153,141 @@ def add_to(
...
+    # If input is a dataset, convert to an RGBA array
+    if isinstance(xx, xr.Dataset):
+        try:
+            xx = xx.odc.to_rgba(bands=bands, vmin=vmin, vmax=vmax)
+        except ValueError as e:
+            raise ValueError(
+                f"Unable to automatically guess RGB colours ({e}). "
+                f"Manually specify bands to plot using the `bands` "
+                "parameter, e.g. `ds.odc.explore(bands=['a', 'b', 'c'])`"
+            ) from e
```
Kirill888: `.add_to` does all of that and does it after all the projection and max size shrinkage. Code here should not perform any colorization and projection change, instead it should call `.add_to` with appropriate parameters.

odc/geo/_map.py
```diff
@@ -153,3 +153,141 @@ def add_to(
...
+    # Create folium Map if required
+    if map is None:
+        map = Map(**map_kwds)
+
+    # If necessary, reproject to map CRS
+    _crs = map_crs(map)
+    layer_name = xx.name  # copy out name as it is lost in reprojection
+    if _crs != xx.odc.crs:
+        xx = xx.odc.reproject(_crs, **reproject_kwds)
+
```
Kirill888: leave this logic to `.add_to`

odc/geo/_map.py
Kirill888: In fact `.add_to(..., map=None)` will give you image data url and map bounds. So if you desire some customization of imagelayers and the like, you can just focus on that.

Code in explore should then do:

1. Check needed libs are installed
   - Ideally raise "actionable" error (Please install ...)
2. Convert Dataset/DataArray to an image+bounds using `.add_to`
    - All projection and colorization happens in there
    - Fill in some defaults that `.add_to` might not know about (picking the right kind of bands for example)
3. Create appropriate layers to add to a map and a map itself if not provided

robbibt: Thanks @Kirill888! Looks like I can simplify this a fair bit - I didn't realise that `add_to` did its own reprojection. I did however reimplement some of the error catching (e.g. too many dims) outside of `.add_to` was that I was finding it would only raise errors _after_ reprojection and loading into memory - which would cause issues if users passed in e.g. hundreds of timesteps instead of one (data would load slowly then only return the error at the end). I'll have another look and see if I was doing something wrong to cause that to happen.

A few quick questions:
1. I would like to be able to support custom resampling when adding to the map, but [`.add_to` doesn't seem to support this](https://github.com/opendatacube/odc-geo/blob/develop/odc/geo/_map.py#L138). Any objection to me adding a `resampling` param to `.add_to`?
2. Does `.add_to` automatically convert `xr.Dataset` inputs to an RGBA array? I can't see `.add_to` as a method on `xr.Dataset`, only [`xr.DataArray`](https://github.com/opendatacube/odc-geo/blob/a9fbfc53625502d14dc7f124470ae0d54eb52f60/odc/geo/_xr_interop.py#L851) (I'd like `.explore()` to work on both: automatically converting and plotting `xr.Dataset` as an multiband RGB image, and  plotting `xr.DataArray`s as a single-band colorised image).
3. What is the advantage of running `.add_to(..., map=None)`, rather than the current approach of simply creating the map first and doing `.add_to(..., map=m)`?

Kirill888: Extra features like resampling config, or more robust input verification should go into `.add_to`.

robbibt: I've worked out the problem with Dask inputs: if you accidently pass in an array with e.g. hundreds of timesteps, it gets loaded into memory inside `compress` here:
https://github.com/opendatacube/odc-geo/blob/develop/odc/geo/_compress.py#L95-L96

But that Dask load happens _before_ the excess time dimension is caught in `_compress_image` here:
https://github.com/opendatacube/odc-geo/blob/develop/odc/geo/_compress.py#L26-L33

That causes a long wait for data to load, with users only receiving the "Expect 2 or 3 dimensional array" error at the very end. Ideally, we'd be able to raise that error much earlier, avoiding the need to wait until the end of a potentially large Dask load.

odc/geo/_map.py
robbibt: Yep, this seems to work well.

odc/geo/_map.py
Kirill888: > I've worked out the problem with Dask inputs: if you accidently pass in an array with e.g. hundreds of timesteps, it gets loaded into memory inside `compress` here: https://github.com/opendatacube/odc-geo/blob/develop/odc/geo/_compress.py#L95-L96

So ideally compress should catch that before calling "compute", something like `_verify_can_compress(xx: xr.DataArray)` that should contain logic similar to:

https://github.com/opendatacube/odc-geo/blob/a9fbfc53625502d14dc7f124470ae0d54eb52f60/odc/geo/_compress.py#L23-L33

which is basically: array `.ndim in (2,3)` after squeezing out single sized dimensions.

Kirill888: > 2\. Does `.add_to` automatically convert `xr.Dataset` inputs to an RGBA array? I can't see `.add_to` as a method on `xr.Dataset`, only [`xr.DataArray`](https://github.com/opendatacube/odc-geo/blob/a9fbfc53625502d14dc7f124470ae0d54eb52f60/odc/geo/_xr_interop.py#L851) (I'd like `.explore()` to work on both: automatically converting and plotting `xr.Dataset` as an multiband RGB image, and  plotting `xr.DataArray`s as a single-band colorised image).

I see, let's keep `.add_to` as is then, and perform `Dataset+bands->DataArray` conversion inside `.explore`.

robbibt: Sounds good, will put something together 👍

robbibt:
@Kirill888 Have made the following changes:

- Simplified `.explore()` by using existing reprojection functionality in `.add_to()`
- Added `_verify_can_compress` to catch unsuitable dims prior to Dask load in `compress`
- Removed error catching from `_compress_image` now those errors are caught upstream
- Added a new `resample` param to `add_to()` to control reprojection resampling method
- Added a more user friendly `ModuleNotFoundError` if Folium isn't installed
- Improved error messages for auto-band name guessing inside `to_rgba()`
- Minor error message wording and documentation improvements

odc/geo/_compress.py
```diff
@@ -19,18 +19,27 @@
...
 
     if im.ndim == 3:
         h, w, nc = im.shape
         bands = np.transpose(im, axes=(2, 0, 1))  # Y,X,B -> B,Y,X
-    elif im.ndim == 2:
+    else:
         (h, w), nc = im.shape, 1
         bands = im.reshape(nc, h, w)
-    else:
-        raise ValueError(f"Expect 2 or 3 dimensional array got: {im.ndim}")
```
Kirill888, line 47: unless code is on the "hot-path", i.e. gets executed millions of times in a tight loop, it's best to keep these kind of checks. Reasons why:

- communicating assumptions of the code to the reader
- so called "defensive programming" that makes code more resilient to future changes or mistakes in other parts of the existing code. Just because right now this is only reachable through another function that performs that check, this won't always stay that way.

Maybe replacing `raise ValueError` with `assert` to indicate that it's the responsibility of the caller to perform such a check, but it's still important to have something that states data shape expectations even within "internal" functions like this one.

Kirill888:
@robbibt thanks, this is great. Let's put back those data shape checks into `._compress` and it's good to merge after that.

odc/geo/_compress.py
robbibt: Thanks for the explanation, makes sense! Will put this back in

robbibt: Done - just left it as a ValueError for now as it felt cleaner that way.

robbibt:
> @robbibt thanks, this is great. Let's put back those data shape checks into `._compress` and it's good to merge after that.

Awesome, thanks for the great feedback. Have made that change, can rebase merge once you approve.

What's the plan for the next `odc-geo` version release? I'd like to get `mask`, `crop` and `explore` into the DEA Sandbox and it would be much easier with a proper release.

Kirill888 (approved):
thanks, good to merge

PR #119, opened 2024-01-25
Saving larger than RAM Dask images to COG on S3

Kirill888:
## What

- Implements `odc.geo.cog.save_cog_with_dask` that saves Dask based xarrays to COG
- Supports local disk and S3 bucket as output
- COG is generated in a "single-pass", overviews are computed and compressed incrementally
- Relies on multi-part-upload functionality of the S3 to write tiles as they are being computed and compressed while keeping COG mandated order of tiles and sub-images
- Uses `imagecodecs` to compress each tile independently as part of Dask graph
- Uses `tifffile` library to create TIFF header that references compressed tiles
- Computes pixel statistics and stores in COG in gdal format to allow quicker open in QGIS over the network
- Supports `YX` (gray), `YXS` (pixel interleaved rgb, rgba) and `SYX` (multi-plane) image modes 

### MPU classes

These really ought to move to a stand-alone lib. The idea here is this: 

- Input is a lazy sequence of byte segments that form a single file (order is important and preserved)
- At graph construction time, sizes of those segments are unknown, some could even be empty
- Neighboring segments are glued together and written to destination using multi-part-upload, keeping resident memory requirements low, while taking care of S3 upload constraints (at least 5Mb chunks except for last one, no more than 10k parts in total)
- Keep track of sizes in bytes for each segment, and any supplied metadata identifying each segment
- Once every segment have been fully realized, call function to generate a header or footer for the data, given now known sizes of each segment (construct TIFF header in our case)
- Flush remaining caches and finalize upload
   - We maintain a buffer of unwritten data of at least 5Mb at the head and tail of file to allow for generating header/footer afterwards while working within the constraints of S3 multi-upload

File sink is implemented to mimic functionality of S3 multi-part upload, it simply writes chunks to temp files, then glues them all together into a final destination, while deleting temp files to keep the total disk requirements close to the output file size, and not doubling it.

## How to review

- All new files are in `odc/geo/cog/`
- `odc/geo/_cog.py` got moved and split into `odc/geo/_rio.py` and `odc/geo/_shared.py`
- changes outside of that are minimal, should probably review those first
- Can probably ignore `_mpu*` and `_s3` files
   - these are responsible to taking compressed tile data and stitching them into COG 

## TODOs

- move all the "MPU" and S3 code into a separate library
- more proper tests
- harmonize Dask/non-dask interfaces and expose both on `.odc.` accessor


Kirill888:
@SpacemanPaul this work has been done a while back, I have now cleaned up the history and rebased against `develop`. While it's not "done done" (see TODOs), I'd like to merge into `develop` as it's a bit of a pain to maintain a separate branch. We also need to make a release for `.explore` and masking functionality that have been added by @robbibt

odc/geo/cog/_tifffile.py
```diff
@@ -0,0 +1,769 @@
...
+    blocksize: Union[Unset, int, List[Union[int, Tuple[int, int]]]] = Unset(),
+    bigtiff: bool = True,
+    overview_resampling: Union[int, str] = "nearest",
+    aws: Optional[Dict[str, Any]] = None,
+    client: Any = None,
+    stats: bool | int = True,
+    **kw,
+) -> Any:
+    """
+    Save cloud optimized geotiff to S3 or file.
```
robbibt: Have made a small suggested update below, but a more detailed docstring intro here would be really useful, ideally explaining how this is different from the other COG writing funcs/methods we have (so users can make an informed choice).

```suggestion
    Save a Cloud Optimized GeoTIFF to S3 or file with Dask.
```

odc/geo/cog/_tifffile.py
```diff
@@ -0,0 +1,769 @@
...
+
+    :param xx: Pixels as :py:class:`xarray.DataArray` backed by Dask
+    :param dst: S3 url or a file path on shared storage
+    :param compression: Compression to use, default is ``DEFLATE``
+    :param level: compression "level", depends on chosen compression
```
robbibt:
```suggestion
    :param level: Compression "level", depends on chosen compression
```

odc/geo/cog/_tifffile.py
```diff
@@ -0,0 +1,769 @@
...
+    :param predictor: TIFF predictor setting
+    :param compressionargs: Any other compression arguments
+    :param overview_resampling: Resampling to use for computing overviews
+    :param blocksize: Configure blocksizes for main and overview images
+    :param bigtiff: Generate BigTIFF by default, set to ``False`` to disable
+    :param aws: Configure AWS write access
+    :param client: Dask client
```
robbibt, line 605: Will the func automatically detect a launched Dask client too, or do we need to pass it in here?

odc/geo/cog/_s3.py
```diff
@@ -0,0 +1,318 @@
...
+    def writer(self, kw, *, client: Any = None) -> PartsWriter:
+        if client is None:
+            client = _dask_client()
+        writer = DelayedS3Writer(self, kw)
+        if client is not None:
+            writer.prep_client(client)
+        return writer
+
+    # pylint: disable=too-many-arguments
+    def upload(
```
robbibt, line 196: Does this only support Dask inputs, or non-Dask COGs too?

robbibt (approved):
This will be an extremely useful feature to the broader EO community, thanks for all your work on this @Kirill888!

I don't have much to contribute on the code itself, although I've left some very minor documentation suggestions and questions on the PR. I have tested this locally myself and it works really well - I was able to compute the Delayed object and write out a pretty large (~8000 x 8000) COG.

I would like to re-iterate how valuable it would be to have this functionality built into the existing `.odc.(write|to)_cog()` methods (even if that meant breaking changes). The biggest risk I can see with this is that users get presented with multiple similar looking tools to do similar tasks, which may contribute to confusion. But happy for this to be merged in now and that work be done at a later date. 🙂

odc/geo/cog/_s3.py
Kirill888: no, this is specifically for Dask, it operates on Dask specific collections of "future" bytes.

odc/geo/cog/_tifffile.py
Kirill888: yes, it will, but rarely there could be several Dask clients.

PR #124, opened 2024-01-30
Unify reproject behaviour on Dataset and DataArray

### LINKED ISSUE #122: Output resolution arguments are dropped in `[Dataset].odc.reproject(...)`
I have a 1 metre LiDAR dataset that I am loading from a raster file via `rioxarray`. I would like to reproject this dataset into a specific CRS (Australian Albers) and resolution (10 metres), similar to using `output_crs` and `resolution` in `datacube.load()`. 

I can easily reproject to the CRS like this:
```
ds.odc.reproject(how="EPSG:3577")
```
However, specifying `resolution` has no effect:
```
ds.odc.reproject(how="EPSG:3577", resolution=10)
```

Although one workaround would be to manually create a new GeoBox with my custom CRS and resolution, I would love to be able to simply set both CRS and resolution directly within `.odc.reproject()` for ease of use - especially since in this example I don't even need a particularly precise output grid (e.g. snapping/alignment), just a quick and dirty transformation to a specific CRS and lower resolution so I can more easily process a very large LIDAR dataset). 

Full code example:
```
import rioxarray
import odc.geo.xr

ds = rioxarray.open_rasterio(
    filename="https://dotazprdauegisextpubst01.blob.core.windows.net/transport-wa-public/bathymetry/rasters/HD20180814_Laser.bag",
    band_as_variable=True,
    chunks={},
)

ds_albers = ds.odc.reproject(how="EPSG:3577")
ds_albers_10m = ds.odc.reproject(how="EPSG:3577", resolution=10)

print(ds_albers.odc.geobox.resolution)
print(ds_albers_10m.odc.geobox.resolution)
```

Kirill888:
Expose all the possible `GeoBox.from_bbox` parameters on `.reproject`, `.output_geobox`,  `.to_crs`

- Closes #122

### DIFF
--- a/odc/geo/_xr_interop.py
+++ b/odc/geo/_xr_interop.py
@@ -594,6 +594,23 @@ def xr_reproject(
     """
     Reproject raster to different projection/resolution.
 
+    :param src:
+      :py:class:`~xarray.Dataset` or :py:class:`~xarray.DataArray` to reproject.
+
+    :param how:
+      How to reproject the raster. Can be a GeoBox or a CRS (e.g. CRS object or
+      an "ESPG:XXXX" string/integer). If a CRS is provided, the output pixel
+      grid can be customised further via ``resolution``, ``shape``, ``tight``,
+      ``anchor``, ``tol``, ``round_resolution``.
+
+    :param resampling:
+      Resampling method to use when reprojecting the raster. Defaults to
+      "nearest", also supports "average", "bilinear", "cubic", "cubic_spline",
+      "lanczos", "mode", "gauss", "max", "min", "med", "q1", "q3".
+
+    :param dst_nodata:
+      Set a custom nodata value for the output resampled raster.
+
     :param resolution:
 
        * "same" use exactly the same resolution as src
--- a/odc/geo/_xr_interop.py
+++ b/odc/geo/_xr_interop.py
@@ -588,6 +588,11 @@ def xr_reproject(
     Reproject raster to different projection/resolution.
 
     This method uses :py:mod:`rasterio`.
+
+    ..seealso::
+
+       :py:meth:`odc.geo.overlap.compute_output_geobox`
+
     """
     if isinstance(src, xarray.DataArray):
         return _xr_reproject_da(
@@ -596,6 +601,15 @@ def xr_reproject(
     return _xr_reproject_ds(
         src, how, resampling=resampling, dst_nodata=dst_nodata, **kw
     )
+
+
+def _extract_output_geobox_params(kw):
+    # NOTE: modifies input, removes keys
+    out = {}
+    for k in ("tight", "anchor", "resolution", "shape", "tol", "round_resolution"):
+        if k in kw:
+            out[k] = kw.pop(k)
+    return out
 
 
 def _xr_reproject_ds(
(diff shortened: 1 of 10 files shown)


robbibt:
Still reviewing this, but a small issue I've found so far: `shape` isn't respected if you try to reproject to the same CRS but different shape:

```
import rioxarray
import odc.geo.xr

ds = rioxarray.open_rasterio(
    filename="https://data.dea.ga.gov.au/derivative/ga_ls_wo_fq_cyear_3/1-6-0/x30/y24/1995--P1Y/ga_ls_wo_fq_cyear_3_x30y24_1995--P1Y_final_frequency.tif",
    band_as_variable=True,
    chunks={},
)

ds_reprojected = ds.odc.reproject(how="EPSG:3577", shape=(100, 100))

print(ds.odc.geobox.shape)
print(ds_reprojected.odc.geobox.shape)
```
image

robbibt:
And a more general comment: I find it difficult to find out what params are actually supported by `.odc.reproject()`. When I try and inspect the available docstring (e.g. shift tab) I get this:
image

...which doesn't provide any mention of the new `GeoBox.from_bbox` params that are now supported (e.g. `resolution`, `anchor`, `shape` etc). 


Is there any way we can get a more detailed/helpful docstring to appear here (or include the new params with default values in the function definition so they are at least visible)? Trying to work out what params are supported by `.odc.()` extension methods is something that other users in DEA have struggled with too - having more descriptive/easy to access docs could really help improve usability here I think.

odc/geo/geobox.py
```diff
@@ -64,12 +64,40 @@
     xy_,
 )
 
-GeoboxAnchor = Union[AnchorEnum, XY[float]]
+GeoboxAnchor = Union[
+    AnchorEnum,
+    XY[float],
+    float,
+    Literal["center"],
```
robbibt, line 71: This is fantastic - will make "anchor" even more usable than the current "align". Is the double "centre" here intended?

Kirill888: yes, both are "valid" spellings

robbibt: Ah missed the spelling difference, makes sense

odc/geo/_xr_interop.py
```diff
@@ -582,13 +583,61 @@ def xr_reproject(
...
+    resolution: Union[SomeResolution, Literal["auto", "fit", "same"]] = "auto",
+    shape: Union[SomeShape, int, None] = None,
+    tight: bool = False,
+    anchor: GeoboxAnchor = AnchorEnum.EDGE,
+    tol: float = 0.01,
+    round_resolution: Union[None, bool, Callable[[float, str], float]] = None,
     **kw,
 ) -> XrT:
     """
     Reproject raster to different projection/resolution.
```
robbibt, line 596:
```suggestion
    :param src:
      :py:class:`~xarray.Dataset` or :py:class:`~xarray.DataArray` to reproject.

    :param how:
      How to reproject the raster. Can be a GeoBox or a CRS (e.g. CRS object or
      an "ESPG:XXXX" string/integer). If a CRS is provided, the output pixel
      grid can be customised further via ``resolution``, ``shape``, ``tight``,
      ``anchor``, ``tol``, ``round_resolution``.

    :param resampling:
      Resampling method to use when reprojecting the raster. Defaults to
      "nearest", also supports "average", "bilinear", "cubic", "cubic_spline",
      "lanczos", "mode", "gauss", "max", "min", "med", "q1", "q3".

    :param dst_nodata:
      Set a custom nodata value for the output resampled raster.
```

robbibt:
Have finished reviewing this - everything is working really nicely. Is there an obvious solution for the `shape` issue above? Let me know if you're planning to fix it before merging this PR, or prefer to merge this as-is and solve that later (in which case I can approve).

Kirill888:
> Have finished reviewing this - everything is working really nicely. Is there an obvious solution for the `shape` issue above? Let me know if you're planning to fix it before merging this PR, or prefer to merge this as-is and solve that later (in which case I can approve).

I'd rather make sure that case has a test and works first. I can merge it without approve anyway, so no issue there.

Thanks for doc suggestions.

Kirill888:
@robbibt this should all be fixed now, I'll merge now.

Issue #125, opened 2024-02-01
Feature suggestion: `.explore()` method for geometry objects and bounding boxes


robbibt:
After sharing the new `.odc.explore()` method within DEA, it was suggested that a similar `.explore()`-style method would be a valuable addition to other `odc-geo` geometry objects too. 

In particular, having an `.explore()` style method available on `Geometry` and `BoundingBox` objects (or even potentially `GeoBoxes` via their extent?) that plots these objects on an interactive map would allow users to:
- Easily verify that the extents/shape of their geometries line up with their expected AOI
- Bugfix issues when creating complex Geometries
- Visually communicate to beginner users where they are about to load or analyse data

It would also add consistency with other geometry-related packages like `GeoPandas` which also provide an `.explore()` method for their spatial data classes.

Kirill888:
I agree that this would be handy, but in my experience I always want to display a combination of different geoboxes, geometries, bounding boxes, and I want syntax for that to be easy to type in the notebook, while still allowing for fine-tuning presentations of individual components. 

On the base level we can start with `explore(..., map=None|folium.Map|ipyleaflet.Map)`, but at user level I'd like a shorter and more "left to right" interface like:

```python
m = folium.Map(....)
gbox1 @ {'color': 'green'} >> gbox2 >> xx.odc.thumbnail(robust=True) >> m
``` 

where `>> m` can be omitted.

Honestly, lack of that functionality is a significant contributor to the absence of "pixel grid documentation", as it's just too hard to visualize all the needed concepts with "raw" matplotlib or SVG.

robbibt:
> I agree that this would be handy, but in my experience I always want to display a combination of different geoboxes, geometries, bounding boxes, and I want syntax for that to be easy to type in the notebook, while still allowing for fine-tuning presentations of individual components.
> 
> On the base level we can start with `explore(..., map=None|folium.Map|ipyleaflet.Map)`, but at user level I'd like a shorter and more "left to right" interface like:
> 
> ```python
> m = folium.Map(....)
> gbox1 @ {'color': 'green'} >> gbox2 >> xx.odc.thumbnail(robust=True) >> m
> ```
> 
> where `>> m` can be omitted.
> 
> Honestly, lack of that functionality is a significant contributor to the absence of "pixel grid documentation", as it's just too hard to visualize all the needed concepts with "raw" matplotlib or SVG.

[...]

PR #128, opened 2024-02-04
Add `.explore()` to GeoBox, Geometry, BoundingBox

robbibt:
This PR addresses #125 by adding `.explore()` methods for interactively plotting `Geometry`, `BoundingBox` and `GeoBox` objects on a Folium map.

This matches the recently added (#104) `.odc.explore()` functionality for exploring `xarray` data, and the existing `.explore()` functionality from `GeoPandas`. It is intended to allow users to:
- Easily verify that the extents/shape of their geometries line up with their expected AOI
- Bugfix issues when creating complex Geometries
- Visually communicate to beginner users where they are about to load or analyse data

GeoPolygon (no issues with complex geometries including MultiPolygons):
image

GeoBox (plots data from the existing `.outline` and `.grid_lines` methods):
image

BoundingBox:
image

Have used the inbuilt `resolution` functionality of the `.geojson()` method to densify features prior to plotting to ensure they are represented correctly on the map (following the UI example here: https://github.com/opendatacube/odc-geo/blob/develop/odc/geo/ui.py#L295-L297).

### DIFF
--- a/odc/geo/geobox.py
+++ b/odc/geo/geobox.py
@@ -7,6 +7,7 @@ import itertools
 import math
 from collections import OrderedDict, namedtuple
 from typing import (
+    Any,
     Callable,
     Dict,
     Iterator,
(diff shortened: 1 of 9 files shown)


odc/geo/geobox.py
```diff
@@ -422,6 +423,55 @@ def qr2sample(
...
+            extent.
+        :param tiles:
+            Map tileset to use for the map basemap. Supports any option
+            supported by :py:mod:`folium`, including "OpenStreetMap",
+            "CartoDB positron", "CartoDB dark_matter" or a custom XYZ URL.
+        :param attr:
+            Map tile attribution; only required if passing custom tile URL.
+        :param map_kwds:
+            Additional keyword arguments to pass to ``folium.Map()``.
+        :param \**kwargs:
```
Kirill888: is this why you need `anomalous-backslash-in-string`?

I think we should just use `kwargs` as param name and not bother with `**` as it interferes with too many things.

Kirill888:
@robbibt let's change `:param \**kwargs:` to just `:param kwargs:` and remove disable for `anomalous-backslash-in-string`. Also if `pylint: disable=` can be moved from just before function to inside the function, that would be ideal.

tests/test_geom.py
```diff
@@ -63,7 +64,8 @@ def test_props():
 
     box1copy = geom.box(10, 10, 30, 30, crs=crs)
     assert box1 == box1copy
-    assert box1.convex_hull == box1copy  # NOTE: this might fail because of point order
+    # NOTE: this might fail because of point order
+    assert box1.convex_hull == box1copy
```
Kirill888, line 68: I often use something like: `(g1^g2).area < tol` in situations like this. That's "XOR" on geometries `g1^g2 <==> (g1-g2)|(g2-g1)`

odc/geo/geobox.py
```diff
@@ -422,6 +423,55 @@ def qr2sample(
...
+            Additional keyword arguments to pass to ``folium.GeoJson``.
```
Kirill888:
```suggestion
            Additional keyword arguments to pass to :py:class:`folium.GeoJson`.
```

Kirill888: With that and after exposing folium docs in here:

https://github.com/opendatacube/odc-geo/blob/37ec5a185e0333eefa06ea766bff32239a55008c/docs/conf.py#L83-L92

we might get click through to the folium docs hopefully, which are here:
https://python-visualization.github.io/folium/latest/

alexgleith:
Looks good to me! I think Kirill has better experience with this type of code than me, and I support his feedback here.

tests/test_geom.py
robbibt: Hmm, I'm not sure where that change came from - I didn't make it intentionally (although I did Black format the entire script...)

tests/test_geom.py
```diff
@@ -933,8 +935,11 @@ def test_deprecation_warnings():
 
 
 def test_filter():
-    _keep = lambda x, y: True
-    _drop = lambda x, y: False
+    def _keep(x, y):
+        return True
+
+    def _drop(x, y):
+        return False
```
robbibt, line 942: Same here - I didn't intentionally make this change although I did Black format.

odc/geo/geobox.py
robbibt: Have made this change in a few places, but it doesn't look like the click-through is working: https://65c0897265175afa7ad50c31--odc-geo-docs.netlify.app/_api/odc.geo.geobox.geobox.explore

Kirill888 (approved):
Thanks

PR #140, opened 2024-04-08
Disallow zero and negative size shapes

### LINKED ISSUE #136: negative size shape is allowed
I noticed that values of `shape` can be negative, this doesn't seem desirable, it has the effect of flipping the transform as expected. 

```python
from odc.geo.geobox import GeoBox
ex = [-120, -100, 35, 55]

g = GeoBox.from_bbox([140, -43, 160, -23],
   "EPSG:4326", 
   shape = [-1, 1])
g
#GeoBox((-1, 1), Affine(20.0, 0.0, 140.0,
#       0.0, 20.0, -60.0), CRS('EPSG:4326'))
```

I only noticed because I wanted to do the GDAL thing of setting one of the dimensions to zero, whereby you only need to use the shape of one side and it helpfully calculates the other from the aspect ratio of the extent. This is nice because it gives a very short heuristic for "about this detailed" that matches how images work, rather than having  to specify a specific resolution or do gymnastics to work out the ratio.   

So, maybe negative shape should be disallowed, and if it's of interest I could PR the `shape = [0, n]` or `shape = [n, 0]`. 

Thanks!

mplough-kobold:
Attempting to create a GeoBox with an x or y dimension of zero results in a zero-division error because it's not possible to calculate the resolution of the box.  Attempting to create a GeoBox with a negative dimension works but flips the transform.  That's confusing so we'd like to disallow it.

This PR raises exceptions when these situations occur; I've also added tests to confirm functionality.

Resolves #136.

### DIFF
--- a/odc/geo/geobox.py
+++ b/odc/geo/geobox.py
@@ -122,6 +122,11 @@ class GeoBoxBase:
         self._extent: Optional[Geometry] = None
         self._lazy_ui = None
 
+        if self.shape.x < 0 or self.shape.y < 0:
+            raise ValueError(
+                f"Got shape {self._shape.yx!r}: negative sizes are not allowed."
+            )
+
     @property
     def width(self) -> int:
         """Width in pixels (nx)."""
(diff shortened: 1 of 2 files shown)


Kirill888 (approved):
thanks

PR #142, opened 2024-04-09
Write band descriptions in save_cog_with_dask

### LINKED ISSUE #138: Feature request: have save_cog_with_dask write band descriptions
Using `write_cog` we get band descriptions, but these don't currently show up if we use `save_cog_with_dask`. Could this be added? For example if I open a COG created with `write_cog` in a text editor I see a block like:
```
<GDALMetadata>
  <Item name="OVR_RESAMPLING_ALG">NEAREST</Item>
  <Item name="DESCRIPTION" sample="0" role="description">green</Item>
  <Item name="DESCRIPTION" sample="1" role="description">red</Item>
  <Item name="DESCRIPTION" sample="2" role="description">nir</Item>
  <Item name="DESCRIPTION" sample="3" role="description">swir1</Item>
  <Item name="DESCRIPTION" sample="4" role="description">swir2</Item>
  <Item name="DESCRIPTION" sample="5" role="description">swir3</Item>
  <Item name="DESCRIPTION" sample="6" role="description">swir4</Item>
  <Item name="DESCRIPTION" sample="7" role="description">swir5</Item>
  <Item name="DESCRIPTION" sample="8" role="description">swir6</Item>
  <Item name="DESCRIPTION" sample="9" role="description">tir1</Item>
  <Item name="DESCRIPTION" sample="10" role="description">tir2</Item>
  <Item name="DESCRIPTION" sample="11" role="description">tir3</Item>
  <Item name="DESCRIPTION" sample="12" role="description">tir4</Item>
  <Item name="DESCRIPTION" sample="13" role="description">tir5</Item>
</GDALMetadata>
```
Which means in QGIS I see band labels like "Band 04: swir1" whereas for a COG made by `save_cog_with_dask` I just see "Band 04"

cc @mplough-kobold @Kirill888

mplough-kobold:
The xarray and rioxarray packages use a dataset's `long_name` attribute to store band names.  This PR saves those band names in COGs written with `save_cog_with_dask`.  The names are written to be as compatible as possible with GDAL's band name writes.

This mechanism may also be used to write other GDAL metadata.

Resolves #138.

### DIFF
--- a/odc/geo/cog/_tifffile.py
+++ b/odc/geo/cog/_tifffile.py
@@ -40,7 +40,7 @@ if TYPE_CHECKING:
 
 
 def _render_gdal_metadata(
-    band_stats: list[dict[str, float]] | dict[str, float],
+    band_stats: list[dict[str, float]] | dict[str, float] | None,
     precision: int = 10,
     pad: int = 0,
     eol: str = "",
@@ -54,6 +54,8 @@ def _render_gdal_metadata(
             ]
         )
 
+    if band_stats is None:
+        band_stats = []
     if isinstance(band_stats, dict):
         band_stats = [band_stats]
 
--- a/odc/geo/cog/_tifffile.py
+++ b/odc/geo/cog/_tifffile.py
@@ -11,6 +11,7 @@ import itertools
 from functools import partial
 from io import BytesIO
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Union
+from xml.sax.saxutils import escape as xml_escape
 
 import numpy as np
 import xarray as xr
@@ -571,6 +572,30 @@ def _norm_compression_tifffile(
 
     predictor = _norm_predictor(predictor, dtype)
     return (predictor, compression, compressionargs)
+
+
+def _gdal_sample_description(sample: int, description: str) -> str:
+    """Make XML line of GDAL metadata.
+
+    :param band: Sample / band number in data array.
+    :param description: Band name in data array.
+
+    :return: GDAL XML metadata line to place in TIFF file.
+    """
+    # GDAL does double escaping; see frmts/gtiff/geotiff.cpp.
+    # We also double escape to maximize compatibility with tools expecting GDAL-generated metadata.
+    double_escaped_description = xml_escape(xml_escape(description))
+    return f'<Item name="DESCRIPTION" sample="{sample}" role="description">{double_escaped_description}</Item>'
+
+
+def _gdal_sample_descriptions(xx: xr.DataArray) -> List[str]:
+    """Translate ``long_name`` attribute (if present) to GDAL sample description metadata.
+
+    :param xx: Pixels as :py:class:`xarray.DataArray` backed by Dask
+
+    :return: List of GDAL XML metadata lines to place in TIFF file.
+    """
+    return [_gdal_sample_description(sample, description) for sample, description in enumerate(xx.attrs.get("long_name", []))]
 
 
 def save_cog_with_dask(
--- a/odc/geo/cog/_tifffile.py
+++ b/odc/geo/cog/_tifffile.py
@@ -597,16 +597,23 @@ def _gdal_sample_description(sample: int, description: str) -> str:
     return f'<Item name="DESCRIPTION" sample="{sample}" role="description">{double_escaped_description}</Item>'
 
 
-def _gdal_sample_descriptions(xx: xr.DataArray) -> List[str]:
-    """Translate ``long_name`` attribute (if present) to GDAL sample description metadata.
+def _band_names(xx: xr.DataArray) -> List[str]:
+    if "band" in xx.coords and xx.coords["band"].dtype.type is np.str_:
+        return list(xx["band"].values)
+    if "long_name" in xx.attrs:
+        long_name = xx.attrs["long_name"]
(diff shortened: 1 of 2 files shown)


odc/geo/cog/_tifffile.py
```diff
@@ -482,7 +490,9 @@ def _patch_hdr(
         if stats is not None:
```
Kirill888: missed `stats=None, gdal_metadata_extra=[...]` case

Kirill888:
thanks for this, we just need to make sure that extra metadata is written even when stats computation is not happening.

Also I'm a bit uncertain about source of description data: attribute `long_name: list[str]`.

1. For YX data `long_name: str` will be a more expected format than `long_name: list[str]`, that's easy enough to fix.
2. I'm not too keen on attributes that apply per element along some dimension, that's what coordinates are for in xarray, but I can see how attribute can be easier to interface with from the user perspective.

odc/geo/cog/_tifffile.py
```diff
@@ -573,6 +583,33 @@ def _gdal_level(compression: str, default=None) -> Any:
...
+    """Translate ``long_name`` attribute (if present) to GDAL sample description metadata.
+
+    :param xx: Pixels as :py:class:`xarray.DataArray` backed by Dask
+
+    :return: List of GDAL XML metadata lines to place in TIFF file.
+    """
+    return [
+        _gdal_sample_description(sample, description)
+        for sample, description in enumerate(xx.attrs.get("long_name", []))
+    ]
```
Kirill888, line 619: Can we make this a bit more flexible on input?

1. Treat `long_name="something"` as equivalent to `long_name=["something"]`
2. If there is `band` coordinate containing strings, use that as the source metadata
3. If there is `long_name` coordinate, use that

It's common to start off with data like `ds: xarray.Dataset` then use something like `xx = ds.to_array('band')` before saving that to multi-sample COG, would be nice if that generated descriptions with original data variable names.

Other note: While I do appreciate documentation comments even on internal functions, function name and type annotations are often enough.

odc/geo/cog/_tifffile.py
```diff
@@ -573,6 +583,33 @@ def _gdal_level(compression: str, default=None) -> Any:
...
+    # GDAL does double escaping; see frmts/gtiff/geotiff.cpp.
+    # We also double escape to maximize compatibility with tools expecting GDAL-generated metadata.
+    double_escaped_description = xml_escape(xml_escape(description))
+    return f'<Item name="DESCRIPTION" sample="{sample}" role="description">{double_escaped_description}</Item>'
+
+
+def _gdal_sample_descriptions(xx: xr.DataArray) -> List[str]:
+    """Translate ``long_name`` attribute (if present) to GDAL sample description metadata.
+
+    :param xx: Pixels as :py:class:`xarray.DataArray` backed by Dask
```
Kirill888: "backed by Dask", is not a requirement here

mplough-kobold:
Thanks for the review, looks like I'll need to do a bit more thinking about the approach.  More to come.

odc/geo/cog/_tifffile.py
mplough-kobold: Thanks, done.

odc/geo/cog/_tifffile.py
mplough-kobold: Thanks, done.

odc/geo/cog/_tifffile.py
mplough-kobold: I added a `_band_names` function above that takes care of all of this.  We now use the `band` coordinate if it's present and it's a string; otherwise we fall back to the `long_name` attribute, and that can be a single string or a list of strings.

mplough-kobold:
> Also I'm a bit uncertain about source of description data: attribute `long_name: list[str]`.

This one was tricky to track down but there are two significant places where it's general practice.  First, [Xarray's documentation](https://docs.xarray.dev/en/latest/getting-started-guide/quick-overview.html#attributes) recommends using it:

> While you’re setting up your DataArray, it’s often a good idea to set metadata attributes. A useful choice is to set `data.attrs['long_name']` and `data.attrs['units']` since xarray will use these, if present, to automatically label your plots. These special names were chosen following the [NetCDF Climate and Forecast (CF) Metadata Conventions](https://cfconventions.org/cf-conventions/cf-conventions.html).

Second, [rioxarray](https://github.com/corteva/rioxarray) writes band names to the `long_name` attribute when opening TIFF files via `rioxarray.open_rasterio` and writing TIFF files via `.  I could not find any specific documentation for which attribute is used, only a [general allusion here](https://corteva.github.io/rioxarray/stable/getting_started/getting_started.html#rioxarray):

> Why use [rioxarray.open_rasterio()](https://corteva.github.io/rioxarray/stable/rioxarray.html#rioxarray.open_rasterio) instead of xarray.open_rasterio?
> ...
> 6. It loads raster metadata into the attributes.

Inspection of attributes after calling that function reveals that `long_name` is used.

---

> I'm not too keen on attributes that apply per element along some dimension, that's what coordinates are for in xarray, but I can see how attribute can be easier to interface with from the user perspective.

Definitely agree with you on this.  I added checks to ensure that the correct number of band names are present if we use `long_name` as the source of our band names.

Kirill888 (approved):
Thanks

PR #146, opened 2024-04-30
Add tag writing support including scale, offset and units special cases

alexgleith:
Added a simple test.

Open to feedback on a better method, but I think this works ok and hopefully handles all cases!

### DIFF
--- a/odc/geo/cog/_rio.py
+++ b/odc/geo/cog/_rio.py
@@ -83,6 +83,40 @@ def _norm_compression_opts(
     return compression
 
 
+def _get_gdal_metadata(
+    xx: xr.DataArray | list[xr.DataArray],
+    tags: Optional[Dict[str, Any]],
+) -> Dict[str, Any]:
+    """
+    Extract metadata from xarray for GDAL.
+
+    :param xx: xarray.DataArray
+    :return: Dictionary of metadata
+    """
+    meta: Dict[str, Any] = {"tags": tags}
+    if not isinstance(xx, list):
+        xx = [xx]
+
+    # The logic for this is to fill the lists with
+    # None values if the attribute is not present
+    # as only some bands may have scale and offset
+    # and we need to know which ones...
+    scales, offsets, units = [], [], []
+    for x in xx:
+        scales.append(x.attrs.get("scales", None))
+        offsets.append(x.attrs.get("offsets", None))
+        units.append(x.attrs.get("units", None))
+
+    if any(v is not None for v in scales):
+        meta["scales"] = scales
+    if any(v is not None for v in offsets):
+        meta["offsets"] = offsets
+    if any(v is not None for v in units):
+        meta["units"] = units
+
+    return meta
+
+
 def _write_cog(
     pix: np.ndarray,
     geobox: GeoBox,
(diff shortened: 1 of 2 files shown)


odc/geo/cog/_rio.py
```diff
@@ -156,17 +157,30 @@ def _write_cog(
     rio_opts.update(extra_rio_opts)
 
     def _write(pix, band, dst):
+        print("Writing")
         if not use_windowed_writes:
```
Kirill888, line 194: Debug leftover

alexgleith: Whoops, that's embarrassing!

odc/geo/cog/_rio.py
```diff
@@ -158,15 +159,27 @@ def _write_cog(
...
+                else:
+                    block = pix[(slice(None),) + win.toslices()]
+
+                dst.write(block, indexes=band, window=win)
+
+        if tags is not None:
+            other_tags = {}
+            for tag, value in tags.items():
+                if tag == "scale":
+                    dst._set_all_scales([value])  # pylint: disable=protected-access
```
Kirill888: There must be an official way to set those, no? Not a fan of this special handling of `scale,offset,units` tags. Those tags should probably be populated from attributes.

alexgleith: From xarray attributes you mean?

I don't know why these are `_` protected access methods, but this is the way Vincent recommended to set them https://github.com/rasterio/rasterio/discussions/2933

Kirill888: Looks like it should be `dst.{scales,offsets,units} = ...`

But, like I mentioned in the above, these should come from attributes, or maybe function parameters not from inside of "tags"

alexgleith: Actually, turns out you can just assign the property like `dst.scales = [0.1]`.

alexgleith: I put them in a `tags` dict to minimise the number of function parameters. If you think it makes sense for them to be handled separately to tags, I can do that. It'll be four parameters instead of one though.

Kirill888: Also should not assume that value is not an array already, with the current implementation there is no way of specifying different values per band.

Another thing to confirm is what exactly happens when one sets `.scales|offsets`, is this just adding metadata  to the output (my assumption), or does this also turn on pixel value conversion.

alexgleith:
So, I checked, and all these tags and special metadata items are written to the GDAL section in the tif metadata:

```
  GDAL Metadata: <GDALMetadata>
  <Item name="random_tag">random_value</Item>
  <Item name="OFFSET" sample="0" role="offset">0.989999999999999991</Item>
  <Item name="SCALE" sample="0" role="scale">100</Item>
  <Item name="UNITTYPE" sample="0" role="unittype">microwaves</Item>
</GDALMetadata>
```

Kirill888:
This is all fine, I do have issue with the name `tags`: since this is TIFF making function, my first assumption is that `tags` refers to TIFF tags and not to "GDAL" tags, aka gdal metadata items, or general "annotation" of some sort. On the other hand [that's the nomenclature used by `rioxarray`](https://github.com/corteva/rioxarray/blob/13a3a99b3779f86e06a5130210a835af937cfa99/rioxarray/raster_writer.py#L88) also, so...

Kirill888:
@alexgleith do you want to rebase this into one clean commit and maybe add a note to docs clarifying that tags refer to "GDAL tags/metadata". As a person who actually read TIFF spec I'm probably biased in my interpretation of `tags`, but it won't hurt being more precise about it.

alexgleith:
Thanks, @Kirill888. I've refactored to simplify the multi-band case a bit.

Can you please check my logic looks like it'll work? I think it's ok...

Kirill888 (commented):
ty

Kirill888 (approved):

PR #151, opened 2024-05-14
Fixes for rotated GeoBox representation on xarray

Kirill888:
- Rotated GeoBox didn't survive serialization to disk, and now does
   - Don't rely on custom `encoding["_transform"]` instead use attribute
- Improved detection of spatial dimensions with non-standard names
   - This change allows packing data arrays with different geo-registration into one `xarray.Dataset`
   - Bands with different native resolutions without upsampling is a good example of such use-case 
- Allow arbitrary names for dimensions in `xr_zeros|wrap_xr` (i.e. `dims=("T", "Y", "X", "B")`)
- Make it easy to force `y,x` names for spatial dimensions even when geobox is in 4326, `always_yx=True`
- Allow arbitrary number of non-spatial dimensions in any position relative to spatial dimensions
- Removed "LayerControl present" detection code in `.odc.explore` as it relies on internal details of `folium.Map`
   - ping: @robbibt, the idea is that if user is supplying existing map to `.explore()` then we won't try adding `LayerControl` or change map view, and just add ImageOverlay to the existing map
   - I had some issues with adding multiple images to the same map, looks like it's due to `LayerControl` and some problem/unexpected behaviour in `folium` that I have not fully debugged
- Mapping functions now expose optional `index=` parameter, used for layer ordering in folium/ipyleaflet

### DIFF
--- a/odc/geo/_map.py
+++ b/odc/geo/_map.py
@@ -1,4 +1,4 @@
-from typing import Any, Optional, Tuple, Dict
+from typing import Any, Dict, Optional, Tuple
 
 import xarray as xr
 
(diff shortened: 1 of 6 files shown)


robbibt (approved):
This looks good to me - I had had some issues with the `LayerControl` code anyway, so this approach is fine.

The rest of the changes look great! I'll definitely use `always_yx=True`.

PR #154, opened 2024-05-16
Fix type hint backward compatibility 

### LINKED ISSUE #153: Use of newer type hinting syntax breaks for Python < 3.10
A recent commit (3b146cbd55279d4ba13930679a53dbbc167b1010) added a function that uses the newer type hinting syntax, specifically the bar (|) operator. This breaks for versions of Python < 3.10. The line that breaks is [here](https://github.com/opendatacube/odc-geo/blame/72e873ad49fe73ad6a9349f9853003dbcb21e181/odc/geo/cog/_rio.py#L87C3-L87C3). `Union` can be used instead to allow backward compatibility.

fbunt:
Fixes #153

### DIFF
--- a/odc/geo/cog/_rio.py
+++ b/odc/geo/cog/_rio.py
@@ -84,7 +84,7 @@ def _norm_compression_opts(
 
 
 def _get_gdal_metadata(
-    xx: xr.DataArray | list[xr.DataArray],
+    xx: Union[xr.DataArray, List[xr.DataArray]],
     tags: Optional[Dict[str, Any]],
 ) -> Dict[str, Any]:
     """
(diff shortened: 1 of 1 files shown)


Kirill888:
@fbunt thanks, but can you please fix this via import from future instead, like here for example:

https://github.com/opendatacube/odc-geo/blob/72e873ad49fe73ad6a9349f9853003dbcb21e181/odc/geo/_xr_interop.py#L8

SpacemanPaul:
I see setup.cfg still advertises support for Python >=3.8.

This should probably be bumped to at least 3.9 as 3.8 is now end-of-lifed and unmaintained.

fbunt:
@Kirill888 Oh I didn't know about that one, thanks! Done.

Kirill888 (approved):

Issue #155, opened 2024-05-21
Apply spatial mask across non-spatial dimensions


colinahill:
`odc.geo.xr.mask` fails for `xarray.DataArray` and `xarray.Dataset` if they have more than the 2 spatial dimensions of latitude and longitude (or equivalent). As other dimensions are common (e.g. time) it would be convenient to allow spatial masking across all other (non-spatial) dimensions.

This also impacts `odc.geo.xr.crop`, which uses `mask`.

robbibt:
Hye @colinahill, that seems problematic - I'm pretty sure I tested this on multi-temporal data originally. Do you have some simple reproducible example I could try this out on?

Edit: we have code below that successfully applies a mask to multitemporal data, so if this no longer works, it's broken recently. What specific error do you get here?
https://knowledge.dea.ga.gov.au/notebooks/How_to_guides/Polygon_drill.html#Mask-data-using-polygon

lymperis-e:
@colinahill I faced a similar problem, with an 'Operands could not be broadcast together' exception message. I setup a dummy test dataset to check if an arbitrary ds could be cropped in case it had extra dimensions, and realized that maybe odc is picky with the dimension order. This was indeed the case, and I was able to crop my original dataset as expected by first transposing `dataset = dataset.transpose(*[d for d in dataset.dims if d not in ("y", "x")] + ["y", "x"])` and then running the crop op `dataset.odc.crop(poly, apply_mask=True)`

PR #156, opened 2024-05-21
Apply spatial mask across non-spatial dimensions

colinahill:
This fixes issue #155

### DIFF
--- a/odc/geo/_xr_interop.py
+++ b/odc/geo/_xr_interop.py
@@ -318,6 +318,17 @@ def mask(
         value_inside=not invert,
     )
 
+    # Repeat rasterized mask across all non-spatial dims
+    sdims = spatial_dims(xx, relaxed=True)
+    non_sdims = {}
+    non_sdims_axis = []
+    for i, (k, v) in enumerate(xx.sizes.items()):
+        if k not in sdims:
+            non_sdims[k] = v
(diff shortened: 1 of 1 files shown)


odc/geo/_xr_interop.py
```diff
@@ -318,6 +318,17 @@ def mask(
         value_inside=not invert,
     )
 
+    # Repeat rasterized mask across all non-spatial dims
```
Kirill888, line 321: Instead of using `spatial_dims` method to get names of spatial dimensions, can you please use `xx.odc.ydim` to obtain location of the y-axis (x-axis always follows y) and expand `rasterized` using that.

```python
non_sdims = (*xx.dims[:ydim], xx.dims[ydim+2:])
```

robbibt:
Hey @colinahill, I'm not sure it should be necessary to loop over the non-spatial dims - theoretically xarray should handle all that automatically when the code runs this `.where()` line: https://github.com/opendatacube/odc-geo/blob/develop/odc/geo/_xr_interop.py#L321-L322

I've unfortunately also not been able to reproduce the issue - running something like this on a multi-temporal dataset successfully applies the mask to all timesteps: 
image

alexgleith:
Closing as it seems it's not necessary and there's been no response from the proposer.

Issue #162, opened 2024-06-17
Investigate `float` image `nan` support


Kirill888:
An issue in `datacube` repo highlighted possible problems around `float` image support in `odc-geo` and `datacube`: https://github.com/opendatacube/datacube-core/issues/1595

When creating `float{32,64}` outputs without explicitly configured `nodata`, `odc-geo` uses floating point `nan` value as a `fill_value/nodata`, but `nodata=nan` attribute is not being set on the data-array itself.

Looks like omitting `nodata=nan` when interacting with GDAL causes issues, at least when generating overviews, however reproject might also be impacted. We don't really have a good test suite for that area of the code currently. One probably needs to set `src_nodata=nan, dst_nodata=nan` for  `float{32,64}` arrays even when `nodata` attributes is not set on the array itself.

It could be that GDAL treats `nodata=None` as an *"assume only valid pixel values are present, do not expect to see `nan`s, so go ahead and use optimized code-path"* directive. So, there could be significant performance gains if you know that data contains only valid pixels and you can communicate that to GDAL. We will need a mechanism to distinguish between *"nodata not supplied, use default behaviour for a given pixel type"* and *"user tells us there are NO missing values in the data at all"*. This can be achieved by swapping `nodata=None` for `nodata="auto"` in those functions that allow such overrides and treat `None` value as a *"all pixels are valid in this data"*, ignore `nodata` attribute even if set on the data.

SpacemanPaul:
Also worth noting (from the core issue linked above) that COGs that do not have `nodata=nan` embedded in their metadata are mishandled by ESRI products (ArcGIS, etc).

It's a separate issue I guess, but ideally whatever we end up doing about GDAL should fix the other too.

Kirill888:
Can confirm that reprojecting `float32` data with `nan` pixels present, does indeed require supplying `nodata=nan` to GDAL.

![image](https://github.com/opendatacube/odc-geo/assets/1428024/083570f3-0258-4fcd-b60b-c6048ce7d64f)

```python
import matplotlib.pyplot as plt
import numpy as np
from odc.geo.geobox import GeoBox
from odc.geo.xr import xr_zeros

gbox = GeoBox.from_bbox(
    (-2_000_000, -5_000_000, 2_250_000, -1_000_000),
    "epsg:3577",
    shape=101,
)
gbox2 = gbox.to_crs(3857)

da = xr_zeros(gbox, dtype="float32") + 1
[...]

PR #163, opened 2024-06-18
refactor: nodata handling #162

Kirill888:
- adding new types
  - ~`MaybeAutoNodata`~ `SomeNodata` -- `None|int|float|str|"auto"`
  - `Nodata` is now `None|float|int`
  - `MaybeNodata` is now ~`None|float|int|str`~  have been removed
  - "auto" replaces what used to be `None`
  - `None` now means "no nodata value"
- `resolve_nodata()` is used to handle nodata options consistently across the library
- `.odc.nodata` is used to extract `nodata` from xarrays
- fixes in reproject/overview generation for float data. It is assumed to have `nan` values, and GDAL needs `nodata=nan` to handle it correctly.

### nodata types explained

- ~`MaybeAutoNodata`~`SomeNodata` should be used in APIs that allow configuring nodata and nodata overrides
- `Nodata` is a type suitable to pass to GDAL `{src,dst)_nodata` it can be `None` which means expect no missing values in the data.
- `FillValue` is a type suitable for `np.full` and is resolved from `Nodata` + `dtype`
   - `NaN` for not-configured floating point images
   - `0` for not-configured integer images
   - value of `nodata` if set (`NaN` is fine here too) 

```
SomeNodata + dtype + [fallback] -> Nodata + dtype -> FillValue
```


SpacemanPaul:
Thanks for the swift action on this Kirill - I will review tomorrow.

robbibt:
Thanks heaps for this @Kirill888 - the test cases appear to match the functionality I was hoping for nicely, but I'll do a "user" test today and verify that TIFFs exported using the updated code work as intended (e.g. in ESRI etc).

Kirill888:
@robbibt thanks, BTW pleas use `pip install odc-geo==0.4.7rc1` for your test (that release is not on conda, only pypi)

robbibt:
I think this is working perfectly. For example, we have a dataset with an Xarray `nodata: nan` attribute:
``` 
import datacube
import odc.geo.xr
from datacube.utils.cog import write_cog

dc = datacube.Datacube()

query_params = dict(
    x=(142.13223, 142.65461),
    y=(-32.17591, -32.54618),
    time=("2022", "2022"),
)

ds = dc.load(product="ga_ls8cls9c_gm_cyear_3", measurements=["edev"], **query_params)
da = ds.edev.squeeze()
```
image


This gets written out with a GeoTIFF nodata flag `nodata=nan` using both `datacube` and `odc-geo` tooling:
```
da.odc.write_cog("nodata_nan_odcgeo.tif")
write_cog(da, "nodata_nan_datacube.tif")
```
image
image


However, if we write out data without a Xarray `nodata: nan` attribute, `datacube` doesn't include a GeoTIFF nodata flag, but now `odc-geo` does!
```
del da.attrs["nodata"]

da.odc.write_cog("nodata_missing_odcgeo.tif")
write_cog(da, "nodata_missing_datacube.tif")
```
image
image

I can though return a true missing nodata value like this:
```
da.odc.write_cog("nodata_truenone_odcgeo.tif", nodata=None)
```
image

odc/geo/types.py
```diff
@@ -21,8 +21,10 @@
 
 MaybeInt = Optional[int]
 MaybeFloat = Optional[float]
-Nodata = Union[float, int, str]
-MaybeNodata = Optional[Nodata]
+FillValue = Union[float, int]
+Nodata = Union[float, int, None]
+MaybeNodata = Union[float, int, str, None]
+MaybeAutoNodata = Union[float, int, str, None, Literal["auto"]]
 T = TypeVar("T")
```
SpacemanPaul, line 27: `MaybeNodata = Nodata | str` seems unexpected as `Maybe` usually means `or None`.  I understand that str might be necessary for cases where `nodata in ('nan', 'NAN', 'NaN')` but I'm not sure why this gets a "Maybe" label?

SpacemanPaul: Oh I see, there's a conversion function in `math.py`

I still think the naming here is confusing but I don't have a good suggestion for a replacement.  Maybe `ApiNodata`?  As in the nodata type the user sees in the API (as opposed the internal nodata value passed to GDAL)?

SpacemanPaul (approved):
I think some of the naming is confusing, but the approach is sound.

robbibt:
I do want to do a quick test of the COG overviews stuff, so will approve this as soon as I've finished that.

robbibt:
> I think some of the naming is confusing, but the approach is sound.

I also find "MaybeAutoNodata" a bit confusing - took me a while to work out what it meant.

odc/geo/types.py
Kirill888: that's more of a compatibility issue, before this `Nodata` could not be `None`, but it could be a `str`, so there was a `MaybeNodata` also. I guess we could delete it assuming nobody was using it.

Kirill888:
> > I think some of the naming is confusing, but the approach is sound.
> 
> I also find "MaybeAutoNodata" a bit confusing - took me a while to work out what it meant.

What about `SomeNodata` instead of `MaybeAutoNodata`, @SpacemanPaul @robbibt

SpacemanPaul:
> > > I think some of the naming is confusing, but the approach is sound.
> > 
> > 
> > I also find "MaybeAutoNodata" a bit confusing - took me a while to work out what it meant.
> 
> What about `SomeNodata` instead of `MaybeAutoNodata`, @SpacemanPaul @robbibt

Yes `SomeNodata` or even `AnyNodata` works for me.

Kirill888:
Updated,

- Removed no longer relevant `MaybeNodata` type, if anyone was using that type, they were probably also using `Nodata` type too, and that have changed, so...

- use `SomeNodata` instead of `MaybeAutoNodata` name
- add `nodata` setter

robbibt:
OK, overview generation is working much better for our DEA Intertidal example: even if `nodata` is not set, we still get a result that follows GDAL's expected functionality:

`datacube-core` with `nodata=nan` Xarray attribute (overviews generated correctly ✔️):


`datacube-core` with no nodata Xarray attribute (overviews generated incorrectly ❌):


`odc-geo` with `nodata=nan` Xarray attribute (overviews generated correctly ✔️):


`odc-geo` with no nodata Xarray attribute (overviews generated correctly ✔️):

Issue #165, opened 2024-06-24
Switch `odc.geo.data.country_geom` to `geodatasets`


Kirill888:
Newest `geopandas>=1.0.0` no longer includes "naturalearth_lowres" dataset used in `odc.geo.country_geom`, recommended substitution is to use `geodatasets` package instead, it deals with fetching and caching of that data.

PR #171, opened 2024-07-11
Update so tests pass with latest versions of upstream libraries.

SpacemanPaul:
1. Directly download naturalearth countries dataset.  (See #165)

     Noting that the country dataset from naturalearth used previously is NOT packaged with geodatasets - so we may as well just hardcode the url and download it directly.

2. Refactor use of `np.find_common_type` (which is not available in Numpy 2.x) to `np.result_type` (which is).

3. Check if the `gcps` metadata field on the xarray is of string type, and if so run `json.loads` over it.  (Required for `rioxarray>=0.15.6` as it now stores the gcps definition as a string)

### DIFF
--- a/odc/geo/_xr_interop.py
+++ b/odc/geo/_xr_interop.py
@@ -210,7 +210,8 @@ def _mk_crs_coord(
     crs_wkt = cf.get("crs_wkt", None) or crs.wkt
 
     if gcps is not None:
-        cf["gcps"] = _gcps_to_json(gcps)
+        # Store as string
+        cf["gcps"] = json.dumps(_gcps_to_json(gcps))
 
     if transform is not None:
         cf["GeoTransform"] = _render_geo_transform(transform, precision=24)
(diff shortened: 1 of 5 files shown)


odc/geo/_xr_interop.py
```diff
@@ -523,13 +524,15 @@ def _extract_gcps(crs_coord: xarray.DataArray) -> Optional[GCPMapping]:
         return None
     crs = _extract_crs(crs_coord)
     try:
+        if isinstance(gcps, str):
+            gcps = json.loads(gcps)
```
Kirill888, line 529: thanks for this one, `rioxarray` switched to using strings for this, we should do the same on the write side, not just read side ⬇️ 

https://github.com/opendatacube/odc-geo/blob/d020f21906952157309af7385f65a9a045528124/odc/geo/_xr_interop.py#L212

Kirill888 (changes requested):
- let's add caching
- let's store gcps as a string like `rioxarray`, now that we can read that format too

Kirill888 (approved):
thanks Paul

PR #178, opened 2024-09-10
fix small typo in _extract_geo_transform()

FlorisCalkoen:
- Fix a small typo in [_extract_geo_transform()](https://github.com/opendatacube/odc-geo/blob/94ad126571cdddcf4884b0f9ea4024dde15d0208/odc/geo/_xr_interop.py#L540). The variable should probably be named `geo_transform_parts` instead of `geo_transfrom_parts`.


robbibt (approved):
Thanks @FlorisCalkoen!

(@SpacemanPaul /@Kirill888 any preference for merge method here?)

PR #180, opened 2024-09-26
Round gridspec tilesize to fix floating point precision issue

alexgleith:
Fixes an issue where small resolutions (i.e., `0.003`) would lead to crazy coordinates in geoboxes for a gridspec, due to floating point issues.

For example, if you create a gridspec like this, the transform has a rounding error with a very small number added at the end:

```python
from odc.geo.gridspec import GridSpec, XY

WGS84GRID30 = GridSpec("EPSG:4326", tile_shape=(5000, 5000), resolution=0.0003, origin=XY(-180, -90))

tile = (50, 50)
geobox = WGS84GRID30.tile_geobox(tile)

geobox.affine
```
> Affine(0.0003, 0.0, -105.00000000000001, 0.0, -0.0003, -13.500000000000014)

The issue is caused when calculating the tile size (the width and height of the tile, in units of the gridspec), which works like this:

```python
from odc.geo.types import shape_, res_

tile_shape = shape_((5000, 5000))
resolution = res_(0.0003)

tile_shape.x * abs(resolution.x)
```

> `1.4999999999999998`

The proposed fix is to round these numbers at `12` decimal places, which is enough to fix this problem and I hope doesn't cause issues in other areas. New test included an all other tests pass.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -78,7 +78,8 @@ jobs:
         run: |
           conda info
           conda list
-          mamba -V
+          # New version of mamba doesn't support -V
+          # mamba -V
           conda config --show-sources
           conda config --show
           printenv | sort
--- a/odc/geo/_dask.py
+++ b/odc/geo/_dask.py
@@ -1,5 +1,5 @@
 from functools import partial
-from typing import Any, Dict, Optional, Sequence, Tuple, Union
+from typing import Any, Dict, Optional, Tuple, Union
 from uuid import uuid4
 
 import dask.array as da
@@ -15,7 +15,7 @@ from .warp import Resampling, _rio_reproject, resampling_s2rio
 
 
 def _do_chunked_reproject(
-    d2s: Dict[Tuple[int, int], Sequence[Tuple[int, int]]],
+    d2s: Dict[Tuple[int, int], list[Tuple[int, int]]],
     src_gbt: GeoboxTiles,
     dst_gbt: GeoboxTiles,
     dst_idx: Tuple[int, int],
--- a/odc/geo/gridspec.py
+++ b/odc/geo/gridspec.py
@@ -66,9 +66,12 @@ class GridSpec:
         self.crs = norm_crs_or_error(crs)
         self._shape = tile_shape
         self.resolution = resolution
+
+        # This is an arbitrary rounding of 12 decimal places
+        # I don't know how to make it more robust
         self.tile_size = xy_(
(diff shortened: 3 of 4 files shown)


Kirill888:
Why?

alexgleith:
>Why?

Why the PR?

See the test output without the fix.

Kirill888 (changes requested):
The proposed solution is not a robust one. Issue it’s trying to solve is not explained in the pr and it feels like the real problem is some place else. Version updates are typically done as a separate commit in this repo

alexgleith:
>The proposed solution is not a robust one. 

I agree. Not sure if there's a better approach, though.

>feels like the real problem is some place else

I looked through the code and debugged pretty deep. This is where the problem is caused.

>Issue it’s trying to solve is not explained in the pr

Ok, I'll explain it. Apologies...

>Version updates are typically done as a separate commit in this repo

Ok, I'll remove the version update.

alexgleith:
Text updated to explain the reasoning @Kirill888.

Kirill888:
Unrelated build fixes should be a separate commit.
There is a robust way of rounding affine matrix already
https://odc-geo.readthedocs.io/en/latest/_api/odc.geo.math.snap_affine.html#odc.geo.math.snap_affine

We can NOT keep adding rounding in random places to support specific issues caused by use of EPSG 4326

omad:
@alexgleith 

Is there an actual problem, other than that the numbers look funny?

alexgleith:
>Is there an actual problem, other than that the numbers look funny?

Yeah, the gridspec drives the input and output grid for an analysis, so the coordinates of the geotiff that's being produced will be screwy too.

alexgleith:
>There is a robust way of rounding affine matrix already

But the fundamental issue is not the affine, it's the GridSpec, which is producing bad GeoBoxes with bad Affines.

And the bad GridSpec is caused by the incorrect rounding on the tile size. It could be fixed after the fact, but that feels like hiding the underlying problem, to me.

>Unrelated build fixes should be a separate commit.

Ok, fair. I can split that noise out.

alexgleith:
So, is there a better way to do this? I'd really like to not need to maintain my own function for creating a gridspec because we can't handle digital numbers here.

This sucks!

image

alexgleith:
Closing. I resolved this using [GeoboxTiles](https://odc-geo.readthedocs.io/en/latest/api.html#geoboxtiles) instead.

I still think this is uncontroversial, but I've moved on!

PR #181, opened 2024-09-30
Expose odc.{crs_coord,grid_mapping}

omad:
- access crs coordinate variable/name


omad:
Already merged, months ago, what was I even thinking...

PR #187, opened 2024-11-20
First draft of top level masking functions for bitmask and enumerated…

alexgleith:
A work in progress, but currently includes functions:

* `odc.geo.masking.bits_to_bool`, which makes a mask using bit flags
* `odc.geo.masking.enum_to_bool`, which makes a mask using `isin` lookups
* `scale_and_offset`, which applies offset and scale values.

Also:

* `odc.geo.masking.mask_clouds` and the two functions below
* `odc.geo.masking.mask_ls`: apply an opinionated mask to landsat c2l2sr data
* `odc.geo.masking.mask_s2`: apply an opinionated mask to sentinel-2 c1l2 data.

I've added odc xarray extensions for them, which I think works. Tested briefly.

So, thoughts? Opinions? Fears and doubts?

### DIFF
--- a/odc/geo/masking.py
+++ b/odc/geo/masking.py
@@ -1,0 +1,340 @@
+# This file is part of the Open Data Cube, see https://opendatacube.org for more information
+#
+# Copyright (c) 2015-2020 ODC Contributors
+# SPDX-License-Identifier: Apache-2.0
+"""
+Functions around supporting cloud masking.
+"""
+
+from typing import Annotated, Any, Callable, Sequence
+import numpy as np
+from xarray import DataArray, Dataset
+
+from enum import Enum
+
+
+class SENTINEL2_L2A_SCL(Enum):
+    """
+    Sentinel-2 Scene Classification Layer (SCL) values.
+    """
+
+    NO_DATA = 0
+    SATURATED_OR_DEFECTIVE = 1
+    DARK_AREA_PIXELS = 2
+    CLOUD_SHADOWS = 3
+    VEGETATION = 4
+    NOT_VEGETATED = 5
+    WATER = 6
+    UNCLASSIFIED = 7
+    CLOUD_MEDIUM_PROBABILITY = 8
+    CLOUD_HIGH_PROBABILITY = 9
+    THIN_CIRRUS = 10
+    SNOW = 11
+
+
+SENTINEL2_L2A_SCALE = 0.0001
+SENTINEL2_L2A_OFFSET = -0.1
+
+
+class LANDSAT_C2L2_PIXEL_QA(Enum):
+    """
+    Landsat Collection 2 Surface Reflectance Pixel Quality values.
+    """
+
+    NO_DATA = 0
+    DILATED_CLOUD = 1
+    CIRRUS = 2
+    CLOUD = 3
+    CLOUD_SHADOW = 4
+    SNOW = 5
+    CLEAR = 6
+    WATER = 7
+    # Not sure how to implement these yet...
+    # CLOUD_CONFIDENCE = [8, 9]
+    # CLOUD_SHADOW_CONFIDENCE = [10, 11]
+    # SNOW_ICE_CONFIDENCE = [12, 13]
+    # CIRRUS_CONFIDENCE = [14, 15]
+
+
+LANDSAT_C2L2_SCALE = 0.0000275
+LANDSAT_C2L2_OFFSET = -0.2
+
+# TODO: QA_RADSAT and QA_AEROSOL for Landsat Collection 2 Surface Reflectance
+
+
+def bits_to_bool(
+    xx: DataArray,
+    bits: Sequence[int] | None = None,
+    bitflags: int | None = None,
+    invert: bool = False,
+) -> DataArray:
+    """
+    Convert integer array into boolean array using bitmasks.
+
+    :param xx: DataArray with integer values
+    :param bits: List of bit positions to convert to a bitflag mask (e.g. [0, 1, 2] -> 0b111)
+    :param bitflags: Integer value with bits set that will be used to extract the boolean mask (e.g. 0b00011000)
+    :param invert: Invert the mask
+    :return: DataArray with boolean values
+    """
+    assert not (
+        bits is None and bitflags is None
+    ), "Either bits or bitflags must be provided"
+    assert not (
+        bits is not None and bitflags is not None
+    ), "Only one of bits or bitflags can be provided"
+
+    if bitflags is None:
+        bitflags = 0
+
+    if bits is not None:
+        for b in bits:
+            bitflags |= 1 << b
+
+    mask = (xx & bitflags) != 0
+
+    if invert:
+        mask = ~mask
+
+    return mask
+
+
+def enum_to_bool(
+    xx: DataArray, values: Sequence[Any], invert: bool = False
+) -> DataArray:
+    """
+    Convert array into boolean array using a list of invalid values.
+
+    :param xx: DataArray with integer values
+    :param values: List of valid values to convert to a boolean mask
+    :param invert: Invert the mask
+    :return: DataArray with boolean values
+    """
+
+    mask = xx.isin(values)
+
+    if invert:
+        mask = ~mask
+
+    return mask
+
+
+def scale_and_offset(
+    xx: DataArray | Dataset,
+    scale: float | None = None,
+    offset: float | None = None,
+    clip: Annotated[Sequence[int | float], 2] | None = None,
+) -> DataArray | Dataset:
+    """
+    Apply scale and offset to the DataArray. Leave scale and offset blank to use
+    the values from the DataArray's attrs.
+
+    :param xx: DataArray with integer values
+    :param scale: Scale factor
+    :param offset: Offset
+    :return: DataArray with scaled and offset values
+    """
+
+    # For the Dataset case, we do this recursively for all variables.
+    if isinstance(xx, Dataset):
+        for var in xx.data_vars:
+            xx[var] = scale_and_offset(xx[var], scale, offset, clip=clip)
+
+        return xx
+
+    # "Scales" and "offsets" is used by GDAL.
+    if scale is None:
+        scale = xx.attrs.get("scales")
+
+    if offset is None:
+        offset = xx.attrs.get("offsets")
+
+    # Catch the case where one is provided and not the other...
+    if scale is None and offset is not None:
+        scale = 1.0
+
+    if offset is None and scale is not None:
+        offset = 0.0
+
+    # Store the nodata values to apply to the result
+    nodata = xx.odc.nodata
+
+    # Stash the attributes
+    attrs = dict(xx.attrs.items())
+
+    if nodata is not None:
+        nodata_mask = xx == nodata
+
+    # If both are missing, we can just return the original array.
+    if scale is not None and offset is not None:
+        xx = (xx * scale) + offset
+
+    if clip is not None:
+        assert len(clip) == 2, "Clip must be a list of two values"
+        xx = xx.clip(clip[0], clip[1])
+
+    # Re-attach nodata
+    if nodata is not None:
+        xx = xx.where(~nodata_mask, other=nodata)
+
+    xx.attrs = attrs  # Not sure if this is required
+
+    return xx
+
+
+# pylint: disable-next=dangerous-default-value
+def mask_invalid_data(
+    xx: DataArray | Dataset,
+    nodata: int | float | None = None,
+    skip_bands: Sequence[str] = [],
+) -> DataArray | Dataset:
+    """
+    Mask out invalid data values.
+
+    :param xx: DataArray
+    :return: DataArray with invalid data values converted to np.nan. Note this will change the dtype to float.
+    """
+    if isinstance(xx, Dataset):
+        for var in xx.data_vars:
+            if var not in skip_bands:
+                xx[var] = mask_invalid_data(xx[var], nodata)
+        return xx
+
+    if nodata is None:
+        nodata = xx.odc.nodata
+
+    assert nodata is not None, "Nodata value must be provided or available in attrs"
+
+    xx = xx.where(xx != nodata)
+    xx.odc.nodata = np.nan
+
+    return xx
+
+
+# pylint: disable-next=too-many-arguments, dangerous-default-value
+def mask_clouds(
+    xx: Dataset,
+    qa_name: str,
+    scale: float,
+    offset: float,
+    clip: tuple,
+    mask_func: Callable = enum_to_bool,  # Pass the function for enum-based masks (bits_to_bool or enum_to_bool)
+    mask_func_args: dict = {},
+    apply_mask: bool = True,
+    keep_qa: bool = False,
+    return_mask: bool = False,
+) -> Dataset:
+    """
+    General cloud masking function for both Landsat and Sentinel-2 products.
+
+    :param xx: Dataset or DataArray
+    :param qa_name: QA band to use for masking
+    :param mask_classes: List of mask class values (e.g., cloud, cloud shadow)
+    :param scale: Scale value for the dataset
+    :param offset: Offset value for the dataset
+    :param clip: Clip range for the data
+    :param includ_cirrus: Whether to include cirrus in the mask
+    :param apply_mask: Apply the cloud mask to the data, erasing data where clouds are present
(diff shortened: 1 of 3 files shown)


Kirill888:
@alexgleith I think that chosen approach for representing "bit query" in `bits_to_bool` is confusing and open to interpretations, in particular the "value part", say you need to check that bit 5 is 1, is value `1<<5` or just `1`.

I'd prefer if we just took a string like this

```python
mask = bits_to_bool(xx, "xx01_1xxx")
```

instead of proposed

```python
mask = bits_to_bool(xx, [3,4,5], b0001_1000)
```

Kirill888:
Other option I guess is a dictionary `int -> 0|1`, so something like

```python
mask = bits_to_bool(xx, {5:0, 4:1, 3:1})
```

but that's way more prone to errors: off by 1 or LSB/MSB confusion.

odc/geo/masking.py
```diff
@@ -0,0 +1,121 @@
...
+    else:
+        if not ignore_missing:
+            raise ValueError(
+                "Scale and offset not provided and not found in attrs.scales and attrs.offset"
+            )
+
+    return xx
+
+
+def scale_and_offset_dataset(
```
Kirill888: this function should not be public, `scale_and_offset` should work on both.

alexgleith: Fair enough, happy to change this.

odc/geo/masking.py
```diff
@@ -0,0 +1,121 @@
...
+        mask = ~mask
+
+    return mask
+
+
+def scale_and_offset(
+    xx: DataArray,
+    scale: float | None,
+    offset: float | None,
+    ignore_missing: bool = False,
```
Kirill888: Without reading code I would assume that `ignore_missing` is about `NaN` handling, but it's actually about attribute handling, I suggest to drop this parameter altogether

odc/geo/masking.py
```diff
@@ -0,0 +1,121 @@
...
+
+    # Catch the case where one is provided and not the other...
+    if scale is None and offset is not None:
+        scale = 1.0
+
+    if offset is None and scale is not None:
+        offset = 0.0
+
+    if scale is not None and offset is not None:
+        xx = xx * scale + offset
```
Kirill888: what about `nodata` handling?

odc/geo/masking.py
```diff
@@ -0,0 +1,121 @@
...
+
+    mask = (xx & bitflags) != 0
+
+    if invert:
+        mask = ~mask
+
+    return mask
+
+
+def enum_to_bool(xx: DataArray, values: list, invert: bool = False) -> DataArray:
```
Kirill888: `values: Sequence[int]` not `: list`, need to be able to accept tuples and sets not just lists.

odc/geo/masking.py
alexgleith: Ok. I feel like it should throw a warning, but agree this is unnecessary.

odc/geo/masking.py
alexgleith: Good point...

alexgleith:
Would be great to get your review @robbibt and perhaps an opinionated DEA Fmask implementation `mask_dea` or `mask_fmask`?

odc/geo/masking.py
```diff
@@ -0,0 +1,340 @@
...
+        SENTINEL2_L2A_SCL.CLOUD_MEDIUM_PROBABILITY.value,
+        SENTINEL2_L2A_SCL.CLOUD_HIGH_PROBABILITY.value,
+        SENTINEL2_L2A_SCL.CLOUD_SHADOWS.value,
+    ]
+    if include_cirrus:
+        mask_values.append(SENTINEL2_L2A_SCL.THIN_CIRRUS.value)
+
+    return mask_clouds(
+        xx=xx,
+        qa_name=qa_name,
```
caitlinadams, line 331: Given that there is a default qa_name, I'm concerned this might fail without clear information to the user if the data they've loaded does not have the `scl` band, and they haven't realised. Should there be checks in the `mask_s2` and `mask_ls` functions that check whether the supplied `qa_name` band is actually available?

odc/geo/masking.py
```diff
@@ -0,0 +1,340 @@
...
+
+    NO_DATA = 0
+    DILATED_CLOUD = 1
+    CIRRUS = 2
+    CLOUD = 3
+    CLOUD_SHADOW = 4
+    SNOW = 5
+    CLEAR = 6
+    WATER = 7
+    # Not sure how to implement these yet...
```
caitlinadams, line 52: I've developed a way to do this in eo-insights: https://github.com/frontiersi/eo-insights/blob/412cfb5e4b416c1cfc7d52286a8b9eb241e730d2/src/eo_insights/masking.py#L267-L268

A two-bit pair can have four values: 0 = 0b00, 1 = 0b01, 2 = 0b10, 3 = 0b11. You can appropriately represent this by inserting the requested integer value at the minimum bit using `integer_value << min(bits)` such that a value of `2` at bits `[2, 3]` for a 4-bit integer produces `0b1000` and a value of `3` at bits `[2, 3]` for a 4-bit integer produces `0b1100`.

odc/geo/masking.py
```diff
@@ -0,0 +1,340 @@
...
+    xx: Dataset,
+    qa_name: str = "scl",
+    include_cirrus: bool = False,
+    apply_mask: bool = True,
+    keep_qa: bool = False,
+    return_mask: bool = False,
+) -> Dataset:
+    """
+    Perform cloud masking for Sentinel-2 L2A products.
+    """
```
caitlinadams, line 319: Is there anything to prevent the user from mistakenly using these functions on inappropriate datasets? What happens if the user has loaded Harmonised Landsat Sentinel-2, which uses fmask, which has a different set of flag definitions (and nodata value) to either of the defaults included here? They would need to know that they can't use either of these functions, and have to use the `mask_clouds` function with custom variables relevant to their product. I'm worried that the inclusion of opinionated masking here is going to trip people up, and there need to be appropriate checks to inform the user that they might be using the function inappropriately.

odc/geo/masking.py
```diff
@@ -0,0 +1,340 @@
...
+    # For the Dataset case, we do this recursively for all variables.
+    if isinstance(xx, Dataset):
+        for var in xx.data_vars:
+            xx[var] = scale_and_offset(xx[var], scale, offset, clip=clip)
+
+        return xx
+
+    # "Scales" and "offsets" is used by GDAL.
+    if scale is None:
+        scale = xx.attrs.get("scales")
```
caitlinadams, line 147: What happens if these attributes aren't available? Should you return `None`?

caitlinadams:
@alexgleith -- is there any chance you could provide a scratch script or notebook that loads the Landsat and Sentinel-2 products you have opinionated masking for, either from a single tiff, or a STAC query? I'd like to run a few tests to understand what might happen if a user applies the opinionated masking functions naively and makes a mistake. You could just send this to me directly if you don't want to have it as part of the repo.

alexgleith:
>is there any chance you could provide a scratch script or notebook that loads the Landsat and Sentinel-2

Here's a notebook: https://gist.github.com/alexgleith/84ebc85e17b904a48319845e42947352

caitlinadams:
Thanks for putting this together, Alex! At this stage, my main concern is that it would be very easy for the user to misuse `mask_ls` and `mask_s2`. Given this, my current preference would be to implement `mask_clouds` as part of odc-geo, and move the opinionated elements to a different form. This could be a repo with importable configurations, or providing examples to users about what values they need to put into `mask_clouds` for different products, or something else we haven't thought of yet!

What do you think?

odc/geo/masking.py
```diff
@@ -0,0 +1,340 @@
...
+LANDSAT_C2L2_SCALE = 0.0000275
+LANDSAT_C2L2_OFFSET = -0.2
+
+# TODO: QA_RADSAT and QA_AEROSOL for Landsat Collection 2 Surface Reflectance
+
+
+def bits_to_bool(
+    xx: DataArray,
+    bits: Sequence[int] | None = None,
+    bitflags: int | None = None,
```
robbibt, line 68: Is it correct that `bits` and `bitflags` are mutually exclusive, e.g. users could provide either option? If so, we should clarify this in the documentation.

odc/geo/masking.py
```diff
@@ -0,0 +1,340 @@
...
+
+def scale_and_offset(
+    xx: DataArray | Dataset,
+    scale: float | None = None,
+    offset: float | None = None,
+    clip: Annotated[Sequence[int | float], 2] | None = None,
+) -> DataArray | Dataset:
+    """
+    Apply scale and offset to the DataArray. Leave scale and offset blank to use
+    the values from the DataArray's attrs.
```
robbibt, line 130: Are these attributes that come with a particular STAC extension? If so, makes me wonder if we should configure these in our DEA datacube products too

odc/geo/masking.py
```diff
@@ -0,0 +1,340 @@
...
+    """
+    General cloud masking function for both Landsat and Sentinel-2 products.
+
+    :param xx: Dataset or DataArray
+    :param qa_name: QA band to use for masking
+    :param mask_classes: List of mask class values (e.g., cloud, cloud shadow)
+    :param scale: Scale value for the dataset
+    :param offset: Offset value for the dataset
+    :param clip: Clip range for the data
+    :param includ_cirrus: Whether to include cirrus in the mask
```
robbibt, line 236: The docstring here doesn't match the current params

odc/geo/masking.py
```diff
@@ -0,0 +1,340 @@
...
+        mask = ~mask
+
+    return mask
+
+
+def scale_and_offset(
+    xx: DataArray | Dataset,
+    scale: float | None = None,
+    offset: float | None = None,
+    clip: Annotated[Sequence[int | float], 2] | None = None,
```
robbibt, line 126: Given this can be applied at the dataset level, we should add a "skip_bands" param here to match

robbibt (changes requested):
### General feedback:

This is brilliant, thanks so much for working on this @alexgleith - getting masking functionality into `odc-geo` is something I've wanted for a long time, and I can see this making many of our current workflows so much easier.

I do think I agree with Caitlin that that we should probably break this PR into two chunks to consider seperately. The generic tooling is "easier" to progress, and is a perfect fit for this repo. However, the opinionated masking is trickier, especially given all the possible edge cases that can occur when users naively start applying the functions to various datasets (e.g. USGS Level 1 Landsat, USGS Level 2, DEA Collection 3, DE Africa, E84, HLS etc). So perhaps, just include the following functions in this initial PR:

- `bits_to_bool`: generic bitmasking
- `enum_to_bool`: generic "is-in" masking
- `scale_and_offset`: generic re-scaling of data by scale/offsets 
- `mask_invalid_data`: generic conversion of integer nodata values to `NaN` (given both this and `scale_and_offset` convert integer data to floating point, I wonder if `mask_invalid_data` should become just a special variant of this function and call `scale_and_offset` under-the-hood?)
- `mask_clouds`: generic framework/workflow for masking clouds from a dataset (e.g. qa band selection, re-scaling, mask creation, mask application

For `mask_clouds` in particular, I think it would be really nice if rather than passing in a custom function, a user could simply provide the bit flags or the enumerated values directly to the function, e.g. perhaps something like this:

```
# Enum masking
ds.odc.mask_clouds(qa_name="fmask", qa_enum=[0, 4], scale=0.0001, offset=0.0)

# Bit masking
ds.odc.mask_clouds(qa_name="pixel_qa", qa_bits=[4, 3], scale=0.0000275, offset=-0.2)
```

That would let `mask_clouds` be used for generic cloud masking tasks and for future satellite datasets with different sets of masking bands and flags than we have seen so far.


### Specific feedback:

1. `.odc.mask_invalid_data` is great - although the first time I ran it, I was suprised that it operated "in-place", e.g. modifying the existing array without having to assign it to a new output.  Is this intended? Both approaches could be useful - I guess we could include an `inplace=False` param in there to allow users to decide how they want to apply it.
    ```   
    ds.odc.mask_invalid_data()
    # vs
    ds_masked = ds.odc.mask_invalid_data()
    ```

2. `.odc.enum_to_bool` works a treat! Again, this one returns a new output instead of modifying data in-place, so we should probably standardise to make sure functionality is expected.

3. I've tested `odc.bits_to_bool` on our WO data, and it works great for masking by individual flags, e.g. `ds.water.isel(time=0).odc.bits_to_bool(bits=[7])` - I actually quite like being able to pass in a simple list of bits. However, I'm not sure how to use it for more complex things like this (which I think relates to @caitlinadams' comments above):

    image

    Bitmasking is something I've found super confusing forever, and I know it's something our users struggle with a lot too. The `datacube.utils.make_mask` function is probably still the most intuitive implementation I've seen, where users can do something like this using human-readable flags. I wonder if we could extent this function to allow something similar here, if variable flags exist in the data, or if a user manually provides them?
    ```
    make_mask(ds.water, water_observed=True)
    ```
4. I tried running `ds.odc.scale_and_offset(scale=0.0001)` on DEA's ARD scaled between 0-10000. It did work, but it also operated in-place which was unexpected, and it appeared to restore nodata values back into the dataset even after converting to float (so the data had values from `-999.0` to `1.0`. I think it's probably a safe assumption that nodata values can be set to `NaN` after rescaling to floating point - I don't think having an integer nodata value is useful at that point.

robbibt:
I guess one other thing to consider here is that there's a bunch of similar masking tools in `odc.algo` which I believe were written to optimise Dask performance: https://github.com/opendatacube/odc-algo/blob/main/odc/algo/_masking.py

We should see if any of that content can be re-used here too.

odc/geo/masking.py
alexgleith: Ok, so if you and Kirill both find this confusing, perhaps it's problematic.

Any suggestions on something better? Improving docs is simple enough... but maybe there's a better way.

odc/geo/masking.py
alexgleith: Yeah, it's fairly standard. I'm not sure of the odc-stac library sets then, but it should.

odc/geo/masking.py
alexgleith: Ok, on it.

odc/geo/masking.py
alexgleith: Good point.

alexgleith:
After chatting with @caitlinadams today, I think I'll suggest that we make the following changes:

- Remove the `mask_ls` and `mask_s2` functions, as they are oversimplified and dangerous
- We can perhaps replace them with dictionaries, which provide opinionated (and editable) config that is passed into the more generic `mask_clouds` function, so that way it can still be easy, like `ds.mask_clouds(**USGS_LS_L2SR)` perhaps.

And a range of code cleanup and doc fixes, plus examples are required.

odc/geo/masking.py
robbibt: Nah, not super confusing, just need to be clear in the docs that a user doesn't need to provide both I think (and most users except the most advanced ones will probably ignore `bitflags` completely).

robbibt:
> After chatting with @caitlinadams today, I think I'll suggest that we make the following changes:
> 
> * Remove the `mask_ls` and `mask_s2` functions, as they are oversimplified and dangerous
> * We can perhaps replace them with dictionaries, which provide opinionated (and editable) config that is passed into the more generic `mask_clouds` function, so that way it can still be easy, like `ds.mask_clouds(**USGS_LS_L2SR)` perhaps.
> 
> And a range of code cleanup and doc fixes, plus examples are required.

I really like the idea of being able to pass in custom dictionary kwarg configs - I think that would allow us to have a really nice generic function that can be highly customisable and support a wide range of current and future sensors.

odc/geo/masking.py
Kirill888: I think `rioxarray` does put them in (it takes those from TIFF tags). Unlike `rioxarray`, `odc-stac` loads from multiple sources that, at least in theory, can have different scale and offset for each asset being loaded, also that information can come from either STAC or from TIFF/netcdf.

There is also this possible confusion of "have these parameters been applied already or not?". 

@robbibt Scale and offset are part of stac raster extension, see

https://odc-stac.readthedocs.io/en/latest/stac-best-practice.html 

in STAC land it means something like "pixels are stored in a quantized form as an integer, to get back to floating point use these scale and offset parameters for all values that are not NODATA, for NODATA use NaN".

PR #190, opened 2024-11-29
Add draft re-write of intro docs

robbibt:
This PR re-writes some of the `odc-geo` intro docs to make it more obvious exactly what the package does, and how it can be used for various geospatial workflows.

I've tried to keep most of the technical info intact, but re-arranged things so we start big-picture and get more detailed as users continue to read.

Very happy to take suggestions - this is intended as a first step that we can continue to improve going forward. 🙂


Kirill888 (approved):
great

robbibt:
@Kirill888 Which of these do you prefer?
image

Kirill888:
@robbibt I prefer `rebase and merge`, gives cleanest history view.

PR #191, opened 2024-12-02
Ensure correct roi precision with numpy 2

Ariana-B:
Numpy 2.0 introduces breaking changes around implicit dtype conversions, causing a loss of precision when computing the roi for  inputs with small resolution values.
See the [original issue](https://github.com/opendatacube/datacube-core/issues/1655) for further details and minimal reproducible example, which uses `datacube.utils.geometry` but produces the same results with the corresponding `odc-geo` methods.


Kirill888 (approved):
Thanks

PR #195, opened 2024-12-13
feat: save COGs with dask to azure

wietzesuijker:
This PR introduces Azure Blob Storage support for saving COGs with Dask, complementing the existing AWS S3 support.

- Implements a MultiPartUpload class in _az.py to handle Azure-specific multipart upload operations, including block staging, finalization, and cancellation.
- Adds MultiPartUploadBase in _multipart.py to provide a shared interface for multipart uploads, applicable to both Azure and S3 implementations.
- Updates save_cog_with_dask to recognize Azure URLs and integrate the Azure multipart upload workflow.
- Adds unit tests in tests/test_az.py using mocked Azure SDK to verify functionality specific to Azure Blob Storage.

Here’s a simple test I ran to verify the Azure implementation:

```python
import dask.distributed
import planetary_computer as pc
from odc.geo.cog._tifffile import save_cog_with_dask
from odc.stac import configure_rio, stac_load
from pystac_client import Client

# Setup Dask client for efficient processing
client = dask.distributed.Client()
configure_rio(cloud_defaults=True, client=client)

# Query STAC API for Sentinel-2 data
catalog = Client.open("https://planetarycomputer.microsoft.com/api/stac/v1")
query = catalog.search(
    collections=["sentinel-2-l2a"],
    datetime="2019-06",
    query={"s2:mgrs_tile": dict(eq="06VVN")},
)
items = list(query.items())
print(f"Found {len(items)} datasets")

# Load Sentinel-2 data with specific bands
resolution = 40 
xx = stac_load(
    items,
    bands=["SCL"],
    resolution=resolution,
    chunks={"x": 2048, "y": 2048},
    patch_url=pc.sign,
    dtype="uint16",
    nodata=0,
)

# Select a single time slice for saving
to_save = xx.SCL.isel(time=3)

# Azure Blob Storage details
blob_name = f"SCL-{to_save.time.dt.strftime('%Y%m%d').item()}.tif"
azure_url = f"az://{container}/{blob_name}"


result = save_cog_with_dask(
    to_save,
    dst=f"az://{container}/{blob_name}",
    compression="DEFLATE",
    predictor=2,
    overview_resampling="nearest",
    azure={
        "account_url": f"https://{storage_account}.blob.core.windows.net",
        "credential": sas_token,  # can be blob specific or container wide
    },
    blocksize=[512, 512],
)

# Compute the result to trigger the actual upload
result.compute()
print(f"COG successfully uploaded to: {azure_url}")
```

While implementing this, I noticed a couple of quirks in the original code:

- Spelling Inconsistencies: Words like finalise vs. finalize were used. I  used **`ise`** rather than `ize`.
- Typing Updates: The code inconsistently uses Dict/List from typing vs. built-ins (dict/list). I've changed it to Dict and List since the dev-env.yml specifies python=3.8 (but I hadn't used python 3.8 in ages).

As this is my first PR here, I look forward to feedback and suggestions for improvement!

### DIFF
--- a/odc/geo/_interop.py
+++ b/odc/geo/_interop.py
@@ -43,6 +43,14 @@ class _LibChecker:
     def tifffile(self) -> bool:
         return self._check("tifffile")
 
+    @property
+    def azure(self) -> bool:
+        return self._check("azure.storage.blob")
+
+    @property
+    def botocore(self) -> bool:
+        return self._check("botocore")
+
     @staticmethod
     def _check(lib_name: str) -> bool:
         return importlib.util.find_spec(lib_name) is not None
--- a/odc/geo/cog/_az.py
+++ b/odc/geo/cog/_az.py
@@ -1,0 +1,191 @@
+import base64
+from typing import Any, Union
+
+from azure.storage.blob import BlobBlock, BlobServiceClient
+from dask.delayed import Delayed
+
+from ._mpu import mpu_write
+from ._multipart import MultiPartUploadBase
+
+
+class AzureLimits:
+    """
+    Common Azure writer settings.
+    """
+
+    @property
+    def min_write_sz(self) -> int:
+        # Azure minimum write size for blocks (default is 4 MiB)
+        return 4 * (1 << 20)
+
+    @property
+    def max_write_sz(self) -> int:
+        # Azure maximum write size for blocks (default is 100 MiB)
+        return 100 * (1 << 20)
+
+    @property
+    def min_part(self) -> int:
+        return 1
+
+    @property
+    def max_part(self) -> int:
+        # Azure supports up to 50,000 blocks per blob
+        return 50_000
+
+
+class MultiPartUpload(AzureLimits, MultiPartUploadBase):
+    def __init__(
+        self, account_url: str, container: str, blob: str, credential: Any = None
+    ):
+        """
+        Initialise Azure multipart upload.
+
+        :param account_url: URL of the Azure storage account.
+        :param container: Name of the container.
+        :param blob: Name of the blob.
+        :param credential: Authentication credentials (e.g., SAS token or key).
+        """
+        self.account_url = account_url
+        self.container = container
+        self.blob = blob
+        self.credential = credential
+
+        # Initialise Azure Blob service client
+        self.blob_service_client = BlobServiceClient(
+            account_url=account_url, credential=credential
+        )
+        self.container_client = self.blob_service_client.get_container_client(container)
+        self.blob_client = self.container_client.get_blob_client(blob)
+
+        self.block_ids: list[str] = []
+
+    def initiate(self, **kwargs) -> str:
+        """
+        Initialise the upload. No-op for Azure.
+        """
+        return "azure-block-upload"
+
+    def write_part(self, part: int, data: bytes) -> dict[str, Any]:
+        """
+        Stage a block in Azure.
+
+        :param part: Part number (unique).
+        :param data: Data for this part.
+        :return: A dictionary containing part information.
+        """
+        block_id = base64.b64encode(f"block-{part}".encode()).decode()
+        self.blob_client.stage_block(block_id=block_id, data=data)
+        self.block_ids.append(block_id)
+        return {"PartNumber": part, "BlockId": block_id}
+
+    def finalise(self, parts: list[dict[str, Any]]) -> str:
+        """
+        Commit the block list to finalise the upload.
+
+        :param parts: List of uploaded parts metadata.
+        :return: The ETag of the finalised blob.
+        """
+        block_list = [BlobBlock(block_id=part["BlockId"]) for part in parts]
+        self.blob_client.commit_block_list(block_list)
+        return self.blob_client.get_blob_properties().etag
+
+    def cancel(self):
+        """
+        Cancel the upload by clearing the block list.
+        """
+        self.block_ids.clear()
+
+    @property
+    def url(self) -> str:
+        """
+        Get the Azure blob URL.
+
+        :return: The full URL of the blob.
+        """
+        return self.blob_client.url
+
+    @property
+    def started(self) -> bool:
+        """
+        Check if any blocks have been staged.
+
+        :return: True if blocks have been staged, False otherwise.
+        """
+        return bool(self.block_ids)
+
+    def writer(self, kw: dict[str, Any], client: Any = None):
+        """
+        Return a stateless writer compatible with Dask.
+        """
+        return DelayedAzureWriter(self, kw)
+
+    def upload(
+        self,
+        chunks: Union["dask.bag.Bag", list["dask.bag.Bag"]],
+        *,
+        mk_header: Any = None,
+        mk_footer: Any = None,
+        user_kw: dict[str, Any] = None,
+        writes_per_chunk: int = 1,
+        spill_sz: int = 20 * (1 << 20),
+        client: Any = None,
+        **kw,
+    ) -> "Delayed":
+        """
+        Upload chunks to Azure Blob Storage with multipart uploads.
+
+        :param chunks: Dask bag of chunks to upload.
+        :param mk_header: Function to create header data.
+        :param mk_footer: Function to create footer data.
+        :param user_kw: User-provided metadata for the upload.
+        :param writes_per_chunk: Number of writes per chunk.
+        :param spill_sz: Spill size for buffering data.
+        :param client: Dask client for distributed execution.
+        :return: A Dask delayed object representing the finalised upload.
+        """
+        write = self.writer(kw, client=client) if spill_sz else None
+        return mpu_write(
+            chunks,
+            write,
+            mk_header=mk_header,
+            mk_footer=mk_footer,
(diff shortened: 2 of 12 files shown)


Kirill888:
@wietzesuijker thanks for the PR.

Can I ask you to undo `dict -> Dict`, `list -> List` changes, we no longer aim to support 3.8 going forward as most of our dependencies do not support it either. This also blows up review effort. When making changes like this please try to separate "clean up changes" from "actual changes" and use separate commits for each separate logical unit of work, example:

- fix spelling
- mypy harmonization
- add doc comments
- implement azure back-end


I'll need to find some time to review this properly and to remember what was the original plan for extending to other blob stores. At this point I'm not convinced we need a base class, but my memory of this code is hazy. In the meantime please

- undo mypy changes
- separate changes to doc comments from the "real change"
- make sure to run `black` formatter, or do the pre-commit thing
- push new change history into this PR with `--force-with-lease` option

odc/geo/cog/_s3.py
```diff
@@ -126,31 +129,33 @@ def write_part(self, part: int, data: SomeData) -> dict[str, Any]:
...
 
     @property
     def url(self) -> str:
+        """Return the S3 URL of the object."""
         return f"s3://{self.bucket}/{self.key}"
 
-    def finalise(self, parts: list[dict[str, Any]]) -> str:
+    def finalise(self, parts: List[Dict[str, Any]]) -> str:
+        """Finalise the multipart upload."""
+        # pylint: disable=import-outside-toplevel,import-error
```
Kirill888: pylint flags?

tests/test_az.py
```diff
@@ -0,0 +1,62 @@
+"""Tests for the Azure MultiPartUpload class."""
+import unittest
+from unittest.mock import MagicMock, patch
+
+from odc.geo.cog._az import MultiPartUpload
+
+
+def test_s3_mpu():
```
Kirill888: test_s3_mpu cut and paste name?

odc/geo/cog/_s3.py
wietzesuijker: oops

tests/test_az.py
wietzesuijker: oops 2

wietzesuijker:
Thanks @Kirill888; I addressed your suggestions.

If you're not too familiar with Azure and want to give it a spin, you can:
- Sign up at [Azure Free](https://azure.microsoft.com/en-us/free/).
- In [Azure Portal](https://portal.azure.com/), create a storage account and container.
- Generate a SAS token in the storage account's Shared access signature settings.
- Use the storage account URL and SAS token in the PR tests.
- Delete the resource group after testing.

Thanks again, and great to have your input on this!

odc/geo/cog/_tifffile.py
```diff
@@ -10,19 +10,22 @@
...
 
 import numpy as np
 import xarray as xr
 
+
 from .._interop import have
 from ..geobox import GeoBox
 from ..math import resolve_nodata
 from ..types import Shape2d, SomeNodata, Unset, shape_
+from ._az import MultiPartUpload as AzMultiPartUpload
```
Kirill888: this will raise an error if `azure` is not installed, we should be able to save to S3 even then though.

Also might as well name that class same way in `_az.py`.

odc/geo/cog/_az.py
```diff
@@ -0,0 +1,191 @@
+import base64
+from typing import Any, Union
+
+from azure.storage.blob import BlobBlock, BlobServiceClient
```
Kirill888: move this import into methods that need that (and tell pylint it's ok). We need to be able to import from here even when `azure` is not installed in the environment.

tests/test_az.py
```diff
@@ -0,0 +1,74 @@
+"""Tests for the Azure MultiPartUpload class."""
+
+import unittest
+from unittest.mock import MagicMock, patch
+
+from odc.geo.cog._az import MultiPartUpload
+
```
Kirill888, line 7: skip this test if `azure` is not available.

odc/geo/cog/_tifffile.py
wietzesuijker: Fair! I've addressed that. I didn't dare to rename the S3MultiPartUpload in _s3.py before. I've also added HAVE_S3 and HAVE_AZURE flags to take care of non-installed packages if either is used. Not sure if you're a fan of that.

tests/test_az.py
wietzesuijker: Added the HAVE_AZURE flag here to be used as decorator @require_azure

odc/geo/cog/_az.py
wietzesuijker: I tried, but I couldn't get `python -m unittest tests/test_az.py` to behave without the import being outside of the class. Would you know how?

Kirill888:
### Optional Dependency Handling

Example:

https://github.com/opendatacube/odc-geo/blob/62b7c5ebe41a3c7fc19f267daf06246e437ff830/odc/geo/_xr_interop.py#L55-L61 

The `have` object is here:

https://github.com/opendatacube/odc-geo/blob/62b7c5ebe41a3c7fc19f267daf06246e437ff830/odc/geo/_interop.py#L22-L28

Add `azure` to that, then use it instead of try imports.

### Tests

We use `pytest` for tests, here is example of skipping whole file when needed dependencies are not present

https://github.com/opendatacube/odc-geo/blob/62b7c5ebe41a3c7fc19f267daf06246e437ff830/tests/test_converters.py#L6-L9

### Packaging

Add optional `azure` feature here:

https://github.com/opendatacube/odc-geo/blob/62b7c5ebe41a3c7fc19f267daf06246e437ff830/setup.cfg#L54-L60


### Run checks locally

`mypy` and `pylint` checks are failing, run them locally or setup actions in your fork, actions on the PR require manual approval  from maintainers.

### Simplifying Review process

There is now quite a bit of cleanup changes here, let's have those in a separate PR that I can merge, then we can focus on actual change without getting bogged down by all the mypy harmonization changes.

wietzesuijker:
Added [a PR](https://github.com/opendatacube/odc-geo/pull/196) for the cleanup. I'll give the rest another push tomorrow (UK time). Hopefully, that will give us enough time to review before the holiday season; if not, we can review again in 2025 :).

Kirill888:
@wietzesuijker other PR is merged, and some other changes, please rebase this against main.

wietzesuijker:
Thanks @Kirill888! I didn't get as much done as I'd liked. I updated the test for az and started using `have.azure` and `have.s3`. Could you review the code to see if I'm on the right track? I might need to git rebase and rewrite the history, but I pushed that issue into tomorrow as I was working on this when it wasn't merged yet.
Thanks again.

odc/geo/_interop.py
```diff
@@ -43,6 +43,14 @@ def datacube(self) -> bool:
...
         return self._check("tifffile")
 
+    @property
+    def azure(self) -> bool:
+        return self._check("azure.storage.blob")
+
+    @property
+    def s3(self) -> bool:
+        return self._check("boto3")
+
```
Kirill888: - we use lower level `botocore` library
- property name is library name, not feature
- `s3 -> botocore` and `boto3 -> botocore`

tests/test_az.py
```diff
@@ -1,96 +1,80 @@
...
-try:
-    from odc.geo.cog._az import AzMultiPartUpload
+import pytest
 
-    HAVE_AZURE = True
-except ImportError:
-    AzMultiPartUpload = None
-    HAVE_AZURE = False
+pytest.importorskip("odc.geo.cog._az")
+from odc.geo.cog._az import AzMultiPartUpload  # noqa: E402
```
Kirill888: This won't work as expected as `odc.geo.cog._az`CAN be imported now even when `azure` is missing, it just won't function after the import. I guess it should be `pytest.importorskip("azure.storage.blob")`

odc/geo/_interop.py
wietzesuijker: fair, changed that to botocore, noting that boto3 is in the requirements.

tests/test_az.py
wietzesuijker: thanks :), changed

Kirill888:
@wietzesuijker there is currently a conflict after merging your other PR, can you please rebase this branch against `develop` branch: `git pull --rebase upstream/develop` or something along these lines.

wietzesuijker:
@Kirill888 I've added some changes to fix non-related mypy errors in https://github.com/opendatacube/odc-geo/pull/195/commits/3b0223c16bbf6459f3f12e39fcbe25b491572b8a and squashed that with your commit.
Also, pylint complained about duplicate code, so I pushed https://github.com/opendatacube/odc-geo/pull/195/commits/e1b6187771bbb1a2413624ef9a43d46c82bbf17f.

The tests pass (after a few tries), and the basic example in the description of the PR still works. 
If you have any other suggestions, let me know.

Kirill888 (approved):
Thanks and sorry for the delay (holidays, and it's summer here too).

PR #206, opened 2025-02-17
feat: overview task batching in save_cog_with_dask

wietzesuijker:
Add tile_batching, tile_batching_threshold, and merge_compressed_overviews for flexible control over tile and overview-layer batching to reduce Dask task overhead and manage chunk limits.

Large COGs with multiple overviews were overloading the Dask scheduler. This change aims to reduce the task count during export, lower scheduler overhead, and improve stability.

### DIFF
--- a/odc/geo/cog/_tifffile.py
+++ b/odc/geo/cog/_tifffile.py
@@ -627,6 +627,9 @@ def save_cog_with_dask(
     blocksize: Union[Unset, int, list[Union[int, tuple[int, int]]]] = Unset(),
     bigtiff: bool = True,
     overview_resampling: Union[int, str] = "nearest",
+    tile_batching: int | Callable[[int], int] = 4,
+    tile_batching_threshold: int = 20,
+    merge_compressed_overviews: Optional[int] = 4,
     aws: Optional[dict[str, Any]] = None,
     azure: Optional[dict[str, Any]] = None,
     client: Any = None,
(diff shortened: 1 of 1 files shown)


wietzesuijker:
@Kirill888 curious to hear your views on this

odc/geo/cog/_tifffile.py
```diff
@@ -716,13 +722,22 @@ def save_cog_with_dask(
             layers[stats].data, nodata=xx_odc.nodata, yaxis=xx_odc.ydim
         )
 
-    # Prepare tiles
+    # Rechunk the input array to match desired COG block sizes,
+    # avoiding a one-to-one mapping of data blocks to tasks.
+    if xx.data.chunksize != meta.chunks:
+        xx = xx.chunk(meta.chunks)
+
```
Kirill888: @wietzesuijker not sure on that one, can you explain why you think this is needed?

Re-chunking can be an expensive operation that can lead to scheduling and memory issues within Dask. If we can work without having one to one chunking regime between input data and COG structure then we should. We can mention in the docs as an optimization strategy to match Dask chunking of input data to that of output COG (best case by construction, rather than rechunking), but I would not force it by default.

User can always perform that rechunk externally if they find that improves performance for their use-case.

odc/geo/cog/_tifffile.py
```diff
@@ -716,13 +722,22 @@ def save_cog_with_dask(
...
+    # Prepare tiles for each pyramid level and band
     _tiles: list["dask.bag.Bag"] = []
     for scale_idx, (mm, img) in enumerate(zip(meta.flatten(), layers)):
         for sample_idx in range(meta.num_planes):
             tt = _compress_tiles(img, mm, scale_idx=scale_idx, sample_idx=sample_idx)
-            if tt.npartitions > 20:
-                tt = tt.repartition(npartitions=tt.npartitions // 4)
+            threshold = max_overview_tasks if max_overview_tasks is not None else 20
```
Kirill888: I'd resolve `max_overview_tasks: int|None` to `max_overviews_tasks: int` at the start of the function as part of "argument parsing step", then just use `max_overview_tasks` without checking for None everywhere.

odc/geo/cog/_tifffile.py
wietzesuijker: Yes, you're right! I've dropped this and added a note to the docstr.

odc/geo/cog/_tifffile.py
wietzesuijker: done

odc/geo/cog/_tifffile.py
```diff
@@ -636,13 +638,19 @@ def save_cog_with_dask(
...
     :param dst: S3, Azure URL, or file path
     :param compression: Compression to use, default is ``DEFLATE``
     :param level: Compression "level", depends on chosen compression
     :param predictor: TIFF predictor setting
     :param compressionargs: Any other compression arguments
     :param overview_resampling: Resampling to use for computing overviews
+    :param overview_batching: Maximum grouping factor for overview tasks to reduce Dask task count
+                              (default 4). Higher values combine more tasks into one.
+    :param max_overview_tasks: Optional cap on the number of Dask tasks per overview level after batching.
+                                If set, no overview level will have more than this many tasks.
```
Kirill888: no, this is not how it used to work.

The hard-coded `20` was really *"only bother with repartition if there are at least 20 tiles in a given image layer"*, it never placed an upper limit on the number compressor tasks. And `4` was *"compress that many adjacent tiles as one task"*. 

I don't think placing an upper limit on the total number of tasks is a good idea anyway, you might reduce total number of tasks (not the way it's currently implemented, because we use Bag.repartition), but you'll have to keep more uncompressed tiles for longer in memory possibly causing blow out.

Also I just now realized this is not "overview" specific, it should really be "tile_batching", this batching is applied to every image layer including native resolution layer.

```python
tile_batching: int = 4,
tile_batching_threshold: int = 20,
```

more likely you'll want `tile_batching` to be a function of the shape of the image layer in tiles or at least a function of the total number of tiles in the layer.

```python
tile_batching: int | Callable[[int], int] = 4
```

something that takes number of tiles in a layer and returns number of tiles to process together, with `1` meaning that repartition don't need to happen at all.

@wietzesuijker Is there an actual usage problem that triggered this work?

wietzesuijker: Thanks @Kirill888, I added both arguments now. Yes, when creating large COGs with many overviews, you create a ridiculous amount of tasks for small jobs, which isn't pleasant to work with for the scheduler, especially if one wants to save multiple large COGs.

odc/geo/cog/_tifffile.py
```diff
@@ -778,12 +796,6 @@ def save_cog_with_dask(
 
     # Upload tiles
     tiles_write_order = _tiles[::-1]  # Reverse tiles for writing
-    if len(tiles_write_order) > 4:  # Optimize for larger datasets
```
Kirill888, line 781: what was the problem with that part? BTW this was a different kind of `4`, it basically was: *"if there are more than 4 image layers, merge 4 highest zoom overviews in compressed domain before spilling to disk"*.

wietzesuijker: Thanks I hadn't fully appreciated that. I still reckon it could be useful if there's many overviews, which is pretty typical for my use case. There are a few other args that I saw like stats, which somewhat arbitrarily picks a level. Also the function has a lot of args now. I saw somewhere the max is set to 15, which this exceeds. Could open an issue for a bigger overhaul if you like.

Kirill888: > saw somewhere the max is set to 15, which this exceeds. Could open an issue for a bigger overhaul if you like.

We break that `15` limit all over the place, so just disabling warning is fine. To me having too many "name-only" parameters that you only use when tuning performance is fine. An alternative would be something like "config class", ideally with a "builder interface" of sorts, but I don't think this fits well here, given it's only an utility method.

odc/geo/cog/_tifffile.py
```diff
@@ -776,14 +816,21 @@ def save_cog_with_dask(
...
         )
 
-    # Upload tiles
-    tiles_write_order = _tiles[::-1]  # Reverse tiles for writing
-    if len(tiles_write_order) > 4:  # Optimize for larger datasets
+    # Reverse tile order for writing: highest overview first
+    tiles_write_order = _tiles[::-1]
+
+    # Optionally merge the top N compressed overviews
+    # to reduce overhead when many overview levels exist.
```
Kirill888: * for S3 output, there is a limit of 10,000 upload chunks per object stored
* Each overview level is an independent *"sub-stream"* (or maybe several, need to check code)
* Each *"sub-stream"* needs to be allocated a subset of the available chunk space ahead of time

This "optimization" is more about reducing number of these "upload sub-streams", so that remaining sub-streams have more breathing room.

wietzesuijker: thanks for explaining :)! I clarified the comment using that

Kirill888 (approved):
Thanks

PR #207, opened 2025-02-21
fix: use math.ceil to force int for repartition

wietzesuijker:
Small fix to use `math.ceil`, as I didn't realize `numpy.ceil` returns float instead of int, which breaks repartitions here


wietzesuijker:
@Kirill888 pardon for the oversight

alexgleith (approved):
Simple, small change. Looks good.

PR #216, opened 2025-03-27
Adding initial support for point cloud data

Kirill888:
## Added support for point cloud data to `odc.geo`

* Xarray Dataset with single dimension and coordinates named `x`, `y` and
    optionally `z` are understood as point cloud data.
* CRS information is exposed via dimensionless `spatial_ref` coordinate
    * same to what `odc.geo` and `rioxarray` use for raster data
* Support CRS information extraction on point data as well as rasters
* Implement Point cloud version of `.odc.reproject(..)`
* Implement `odc.io.load_las` method

## Changes to `odc.geo.xr`

* Implemented normalized access to spatial coordinates via `.odc.{x,y[,z]}: xr.DataArray|None` properties
* Expose `xr_crs_coord` method for constructing coord for CRS metadata

## Other

* fix test that used to write test tiff into a current working directory
* Expose `pyproj.Transformer` creation on a `CRS` object (`crs.transformer(4326) -> pyproj.Transformer`)


Kirill888:
please don't merge just yet

Kirill888:
@omad it's good to go for review, but I might have to merge it soonish, i.e. before the week is over, as I need a new version published.

PR #217, opened 2025-03-28
Fix multipart upload

wietzesuijker:
This PR optimizes Azure multipart uploads by ensuring that only non-empty, strictly ordered block IDs are committed.

- Fixed block IDs: Changed block IDs in AzMultiPartUpload.write_part to use a zero‐padded format (block-{part:05d}) so that all IDs have uniform length.
- Sorted block list: In AzMultiPartUpload.finalise, the parts list is now sorted by PartNumber before creating the BlobBlock list, ensuring correct ordering.
- Improved merge logic: In MPUChunk.merge, the new nextPartId is set to one greater than the maximum PartNumber in the merged parts, ensuring monotonically increasing part numbers.
- Final flush handling: In MPUChunk.flush, when flushing leftover data in left_data, if finalising and the leftover is smaller than write.min_write_sz, the block is still written without asserting its size.
- Data Conversion: In MPUChunk.maybe_write, the spill data is explicitly converted to bytes before being passed to the writer.

However, this error still persists: HttpResponseError: The specified block list is invalid / ErrorCode:InvalidBlockList.

Next steps
I suspect a 0‐byte block (or an out‐of‐order/duplicate block) is still being staged. I will try things along these lines next:
- Adjust maybe_write to skip writing if the spill data is empty.
- Verify that all part numbers are strictly increasing across merges.
- Ensure that no empty block (i.e. 0 bytes) ever gets sent to Azure.

### DIFF
--- a/odc/geo/cog/_az.py
+++ b/odc/geo/cog/_az.py
@@ -1,37 +1,73 @@
 import base64
+import sys
+import threading
+import time
+import uuid
 from typing import Any
+from dask.distributed import get_client, Lock
+from dask.base import tokenize
 
+from azure.core.exceptions import AzureError, HttpResponseError
+from azure.storage.blob import BlobBlock, BlobServiceClient, ContentSettings
 from ._multipart import MultiPartUploadBase
+
+import logging
+
+logger = logging.getLogger("odc.geo.cog._az")
+# Set default level (can be overridden by application config)
+logger.setLevel(logging.INFO)
+# Avoid adding handlers multiple times if this module is reloaded
+if not logger.hasHandlers():
+    # Configure console handler
+    handler = logging.StreamHandler(sys.stdout)
+    formatter = logging.Formatter(
+        "%(asctime)s [%(levelname)s] %(name)s (%(funcName)s): %(message)s"
+    )
+    handler.setFormatter(formatter)
+    logger.addHandler(handler)
+    # Prevent logs from propagating to the root logger if handlers are added here
+    logger.propagate = False
+
+# Example: Increase log level for debugging
+# logger.setLevel(logging.DEBUG)
 
 
 class AzureLimits:
     """
     Common Azure writer settings.
+    See: https://learn.microsoft.com/en-us/rest/api/storageservices/put-block#remarks
+    See: https://learn.microsoft.com/en-us/rest/api/storageservices/put-block-list#remarks
     """
 
     @property
     def min_write_sz(self) -> int:
-        # Azure minimum write size for blocks (default is 4 MiB)
-        return 4 * (1 << 20)
+        # While Azure allows smaller blocks, larger blocks are more efficient.
+        # Let _mpu handle buffering; min practical size here isn't strictly enforced by API.
+        # Setting to a reasonable value like 4MiB helps _mpu manage flushes.
+        return 4 * (1 << 20)  # 4 MiB
 
     @property
     def max_write_sz(self) -> int:
-        # Azure maximum write size for blocks (default is 100 MiB)
-        return 100 * (1 << 20)
+        # Max size for Put Block is 100 MiB for most service versions
+        # Newer versions support 4000 MiB, but stick to 100 MiB for compatibility
+        return 100 * (1 << 20)  # 100 MiB
 
     @property
     def min_part(self) -> int:
-        return 1
+        # Block IDs are user-defined strings, no numeric requirement like S3 parts
+        return 1  # Used conceptually by _mpu
 
     @property
     def max_part(self) -> int:
-        # Azure supports up to 50,000 blocks per blob
+        # Max number of blocks per blob is 50,000
         return 50_000
 
 
 class AzMultiPartUpload(AzureLimits, MultiPartUploadBase):
     """
-    Azure Blob Storage multipart upload.
+    Azure Blob Storage multipart upload using Put Block and Put Block List.
+
+    Manages staging blocks and committing them. Designed for use with Dask.
     """
 
     # pylint: disable=too-many-instance-attributes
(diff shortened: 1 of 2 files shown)


wietzesuijker:
@Kirill888 fyi, WiP after #212

Kirill888:
@wietzesuijker I don't think this is correct way to approach this, looks like we are trying to debug by code change, no good.

Need proper test suite for chunk merging and it's interaction with chunk flushing.

odc/geo/cog/_mpu.py
```diff
@@ -240,9 +238,13 @@ def flush(
...
         if self.left_data:
-            assert len(self.left_data) >= write.min_write_sz
             partId = 1 if leftPartId is None else leftPartId
-            self.parts.insert(0, write(partId, self.left_data))
+            # If finalising, allow flushing leftover even if smaller than min_write_sz.
+            if finalise and len(self.left_data) < write.min_write_sz:
+                self.parts.insert(0, write(partId, self.left_data))
+            else:
+                assert len(self.left_data) >= write.min_write_sz
+                self.parts.insert(0, write(partId, self.left_data))
```
Kirill888: NO, can't do that. The assumption is that if you DO keep data on the left, it must be at least as large as a minimal size allowed by the storage backend.

If you are finalizing and you have unflushed left side data, that data needs to be large enough to allow write to the reserved first block, there should not have been any writes to `partId=1` block prior to this.

Kirill888: After every merge the following should hold:

- nothing was flushed yet, all data is in memory
- OR
- parts have been written out, and `.left_data` is large enough to be flushed without any padding.

PRIOR to merge assumptions are:

- left side has at least one write credit left so that it can flush `rhs.left_data`

If you are running out of write credits during merge that's already a failure that can not be repaired. I guess we need to make sure to NOT use up last credit in normal operation, if only one write is left then DO NOT flush even if you have enough bytes to make a write, instead just keep appending to in-memory storage and flush it along with `rhs.left_data` when merge happens.

wietzesuijker:
@Kirill888 Thanks! I added quite some debugging to find out how to make this work. With this code, I wrote a set of sizeable COGs to ABS without errors as quick as you'd expect it to be. The style of the code is currently still a bit out of sync with the rest of the library. I'm curious to get your thoughts on the implementation, though.

- Block IDs: Uses unique, Base64-encoded block IDs containing a zero-padded part number (e.g., prefix-p000001).
- Part Numbering: Assigns part numbers sequentially. During merges, the next part number is based on the highest number found in the combined parts list, using a Dask lock for safety.
- Final Commit: Builds the final block list by querying Azure for uncommitted blocks matching the upload's prefix, decoding part numbers from the block IDs, sorting numerically, and then committing that sorted list.
- Data Handling: Ensures data is converted to bytes before being written.

odc/geo/_version.py
```diff
@@ -1,3 +1,3 @@
 """version information only."""
 
-__version__ = "0.4.10"
```
Kirill888: please revert, version is handled differently now

Kirill888:
@wietzesuijker it's too much for me to review in a hurry, but we certainly don't need global locking during merge or partid sorting, they MUST be sorted by construction, see my comments from few days back.

odc/geo/cog/_az.py
```diff
@@ -1,37 +1,73 @@
...
+        "%(asctime)s [%(levelname)s] %(name)s (%(funcName)s): %(message)s"
+    )
+    handler.setFormatter(formatter)
+    logger.addHandler(handler)
+    # Prevent logs from propagating to the root logger if handlers are added here
+    logger.propagate = False
+
+# Example: Increase log level for debugging
+# logger.setLevel(logging.DEBUG)
+
```
Kirill888, line 34: no logging setup like that on import please, do it inside your test app if needed for debugging

odc/geo/cog/_mpu.py
```diff
@@ -240,46 +394,131 @@ def flush(
...
+            if len(self.left_data) >= write.min_write_sz:
+                partId = self.nextPartId if leftPartId is None else leftPartId
+                self.parts.insert(0, write(partId, self.left_data))
+                bytes_written += len(self.left_data)
+                self.left_data = bytearray()
+            else:
+                logging.getLogger(__name__).debug(
+                    "Leftover data size (%s bytes) is less than min_write_sz; deferring flush.",
+                    len(self.left_data),
+                )
```
Kirill888, line 406: NO: it's non-recoverable error and a sign of either misconfiguration or mistake in code

odc/geo/cog/_mpu.py
```diff
@@ -240,46 +394,131 @@ def flush(
...
         if finalise:
+            for part in self.parts:
+                if "PartNumber" not in part or "BlockId" not in part:
+                    raise ValueError(f"Malformed part metadata: {part}")
```
Kirill888, line 410: NO, it's upto impl to decide what is malformed or not.

odc/geo/cog/_mpu.py
```diff
@@ -86,6 +92,7 @@ def __init__(
         self.observed: list[tuple[int, Any]] = [] if observed is None else observed
         self.is_final = is_final
         self.lhs_keep = lhs_keep
+        self._global_counter = 0
```
Kirill888, line 95: NO, can't have that

odc/geo/cog/_mpu.py
```diff
@@ -145,93 +151,241 @@ def merge(
...
+            new_nextPartId,
             rhs.write_credits,
             rhs.data,
             lhs.left_data,
-            lhs.parts + rhs.parts,
+            all_parts,
             lhs.observed + rhs.observed,
             rhs.is_final,
             lhs.lhs_keep,
         )
```
Kirill888, line 174: NO, I commented on this before, parts must be ordered by construction, and this logic for next part selection is bogus.

Kirill888: and no lock is needed, nor will this work in a distributed setting

Kirill888:
this is too much change and it contradicts original design, I think there is misunderstanding on how this is supposed to work.

wietzesuijker:
Thanks, I've opened a clean alternative in https://github.com/opendatacube/odc-geo/pull/223

PR #219, opened 2025-04-03
Use pyproject+flit for building package

### LINKED ISSUE #215: Migrate to `pyproject.toml` fully to match other odc projects
We have `odc-loader` and `odc-stac` using new and improved way of defining python module metadata using `pyproject.toml` with `hatchling` build backend, same migration should happen to `odc-geo`.

omad:
I haven't switched to a [`src` layout](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/), whereas [odc-loader](https://github.com/opendatacube/odc-loader/) is using a src-layout.

None of the other ODC repos use a src layout.

Closes: #215


Kirill888 (approved):
great, thanks

Issue #224, opened 2025-04-11
Not an ideal solution of enforcing a different default


emmaai:
I have read the background info and issues with default `XSCALE` and `YSCALE` on gdal.
I'd think this is an equally "less ideal" solution to gdal default, as some applications might be well ok with gdal default computation. 
Wondering if this could be optional in higher levels, instead of enforcing here. I'm not sure if I understand the points here: gdal provides the options either way, default vs specifics, the function in the code totally takes away this option. In my understanding, if applications calling this function face troubles or issues, they can well send in the parameters in `kwargs`.

https://github.com/opendatacube/odc-geo/blob/33396d951d1d7d0e293c81f3a2ece3c2c7307bd8/odc/geo/warp.py#L176

robbibt:
Not sure if you've seen these @emmaai, but there's some more info here - @SpacemanPaul and @Kirill888 did some in-depth exploration of this issue a few years ago:
https://github.com/opendatacube/datacube-core/issues/1448
https://github.com/opendatacube/datacube-core/pull/1450
https://github.com/opendatacube/datacube-core/pull/1457

emmaai:
Yes, I read through the posts. I have no issue with the solution of passing `XSCALE` and `YSCALE` to resolve image distortion, but I'm not sure it's a good idea to almost "hard-code" it in the lowest level of function. As I cited, there might be user case not favouring `XSCALE=1` and `YSCALE=1` but computation results by gdal. The code is basically to force any application to perform gdal equivalent computation before calling the function, if `XSCALE/YSCALE !=1`.

Kirill888:
@emmaai read c++ code from gdal where this is used, it might change your mind

Kirill888:
Quote from the author of gdal:

> Those are undocumented options for now, mostly because this is an area not judged fully satisfactory and they are considered as temporary means pending finding something better.

emmaai:
> Quote from the author of gdal:
> 
> > Those are undocumented options for now, mostly because this is an area not judged fully satisfactory and they are considered as temporary means pending finding something better.

I think it's documented now. https://gdal.org/en/stable/doxygen/structGDALWarpOptions.html#a0ed77f9917bb96c7a9aabd73d4d06e08

If I understand this correctly, the catch is that the `XSCALE` and `YSCALE` are arbitrary, they will change if chunk sizes change. Thus, even the input data are the same, when chunking differently (including padding the edge differently) the final results would be different.

[...]

PR #225, opened 2025-04-15
Stop the parameters with `None` from sending to gdal

emmaai:
As the title, to close issue #224, and re-enable the option of using the default computation in `gdal`. It doesn't affect the behaviour expected by others using this function.


Kirill888 (approved):
Thanks

PR #226, opened 2025-04-15
geom: add type signature

pjonsson:


odc/geo/geobox.py
```diff
@@ -235,7 +235,7 @@ def extent(self) -> Geometry:
         if self.linear:
             _extent = geom.polygon_from_transform(self._shape, self._affine, self._crs)
         else:
-            _extent = geom.polygon(self.boundary(16).tolist(), self._crs).transform(
+            _extent = geom.polygon(self.boundary(16).tolist(), self._crs).transform(  # type: ignore[arg-type]
```
pjonsson, line 238: I don't know if MyPy is correct, but it believes this is a string. Ignore the type error here so `geom.polygon()` can get a type annotation.

Kirill888: I think it's probably due to `numpy.ndarray` typing changes.

Kirill888 (approved):

PR #227, opened 2025-04-21
Add some more type signatures

pjonsson:


odc/geo/ui.py
```diff
@@ -84,7 +84,7 @@ def svg_base_map(
     sz=360,
     map_svg_path=None,
     target=None,
-):
```
Kirill888, line 87: Since we are making this typed function, lets add input side types also. Same applies for function below `make_svg`

scripts/patch_version.py
```diff
@@ -33,14 +33,14 @@ def patch_version_lines(lines, build_number):
         yield line
 
 
-def patch_file(fname, build_number):
```
Kirill888: this is some CI only script, not sure we need types here.

odc/geo/roi.py
```diff
@@ -77,7 +77,7 @@ def chunks(self) -> Chunks2d: ...
 
     def locate(self, pix: SomeIndex2d) -> Tuple[int, int]: ...
 
-    def __dask_tokenize__(self): ...
+    def __dask_tokenize__(self) -> None: ...
```
Kirill888: pretty sure this function is supposed to return something

odc/geo/crs.py
```diff
@@ -122,8 +122,8 @@ def __init__(self, crs_spec: Any):
     def __getstate__(self):
         return {"crs_str": self._str}
 
-    def __setstate__(self, state):
-        self.__init__(state["crs_str"])
+    def __setstate__(self, state) -> None:
+        self.__init__(state["crs_str"])  # type: ignore[misc]
```
Kirill888: I guess mypy wants us to do something like:

```suggestion
    def __setstate__(self, state) -> None:
        self._crs, self._str, self._epsg = _make_crs(state["crs_str"])
```

odc/geo/roi.py
Kirill888: https://github.com/dask/dask/blob/3a810721c0d6bba760cbf472f9ae907e809de92c/dask/typing.py#L215-L218

```python
    @abc.abstractmethod
    def __dask_tokenize__(self) -> Hashable:
        """Value that must fully represent the object."""
        raise NotImplementedError("Inheriting class must implement this method.")
```

Kirill888 (changes requested):
see comments for what needs fixing

odc/geo/ui.py
pjonsson: Fixed.

odc/geo/roi.py
pjonsson: I've updated the return type, but I would have expected an `@override` on the method so I think there's something more that should be fixed in this area. That is a separate issue though, I have no idea how these things are supposed to be stringed together.

scripts/patch_version.py
pjonsson: I don't think it costs much to have the types that are easy though, and having them enables at least some checking from MyPy.

odc/geo/crs.py
pjonsson: Nice!

odc/geo/roi.py
Kirill888: Does `@override` even makes sense on a `Protocol` type? It's really just formalizing duck-typing expected by your code.

scripts/patch_version.py
Kirill888: what's the point of return only type annotation on a file only used on CI? You are changing script that has NO types at all, to a script with VERY PARTIAL types, to me this is worse than the prior state. Please revert this part.

odc/geo/roi.py
pjonsson: I don't know anything about dask, but the `DaskCollection` you linked has bothered to make the method an abstract method through abc, so it's pretty clear that they intend subclasses to override and implement that method along with the other abstract methods.

Edit: I don't think this PR makes the situation worse with respect to overriding the abstract methods though, it was an off-the-cuff comment about the probable intention of dask since we were looking at that piece of code.

tests/test_mpu.py
```diff
@@ -72,7 +72,7 @@ def _mk_fake_data(sz: int) -> bytes:
     return np.random.bytes(sz)
 
 
-def _split(data: bytes, offsets=List[int]) -> Tuple[bytes, ...]:
+def _split(data: bytes, offsets: List[int]) -> Tuple[bytes, ...]:
```
pjonsson, line 75: I assume this was what was meant by the previous assignment.

scripts/patch_version.py
pjonsson: Fixed.

Kirill888 (approved):

PR #234, opened 2025-06-29
Fix GEOSException when reprojecting antimeridian-spanning geometries

egagli:
Fixes GEOSException when reprojecting data in sinusoidal projection that spans the antimeridian

**`odc/geo/geobox.py`:**
- Enhanced `GeoboxTiles.grid_intersect()` with error handling for antimeridian-spanning geometries
- Added `wrapdateline` parameter to `GeoBox.footprint()` method for antimeridian-aware geometry conversion

**`tests/antimeridian.py`:**
-Added antimeridian scenario

I tried the scenario in issue #147 and it seems to work? But couldn't figure out how to make it into a test, or if that's even desirable.

### DIFF
--- a/odc/geo/geobox.py
+++ b/odc/geo/geobox.py
@@ -253,7 +253,11 @@ class GeoBoxBase:
         return span / npoints
 
     def footprint(
-        self, crs: SomeCRS, buffer: float = 0, npoints: int = 100, wrapdateline: bool = False
+        self,
+        crs: SomeCRS,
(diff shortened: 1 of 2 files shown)


SpacemanPaul (changes requested):
Thanks, this looks useful.  

If you can address the issues picked up by pylint, black and mypy, I'd be happy to merge.

SpacemanPaul (approved):

Issue #240, opened 2025-08-05
default simplify value in odc.geo.geom.Geometry.geojson results in misleading display for small shapes


MarkusZehner:
Using two shapes as an input and plotting their intersection via the explore function shows them simplified to the point of being misleading, 
which is fixed when setting [geom.py#L889](https://github.com/opendatacube/odc-geo/blob/9804d5bb21aae6f58ab35c2b0efedc5e9667adde/odc/geo/geom.py#L889) to geojson = self.geojson(resolution=res,  simplify=0).
While a quick way to explore possibly large vectors is really nice, this may be confusing when exploring tiny shapes. I would not expect the default to remove 1/4 of the edges. 

```python
import folium
from odc.geo.geom import Geometry


request = {
    'type': 'Polygon', 
    'coordinates': (
        (
            (3.8809034907597493, 13.562874251497007), 
            (3.8809034907597493, 13.607784431137725), 
            (3.9271047227926035, 13.607784431137725), 
            (3.9271047227926035, 13.562874251497007), 
            (3.8809034907597493, 13.562874251497007),
        ),)
    }

scene = {
    'type': 'Polygon', 
    'coordinates': (
        (
            (3.9243246, 13.5788296), 
            (3.9448982, 13.6706209), 
            (3.9783441, 13.8192305), 
            (4.0121515, 13.9678991), 
            (4.0459251, 14.1166378), 
            (4.0794668, 14.2652085), 
            (4.1127214, 14.4136498), 
            (4.125277, 14.4697949), 
            (4.9464047, 14.4647797), 
            (4.9380873, 13.4725398), 
            (3.9239229, 13.4783177), 
            (3.9243246, 13.5788296),
        ),)
    }
request = Geometry(request, crs = 4326)
scene = Geometry(scene, crs = 4326)

m = folium.Map()
m = request.explore(m)
m = scene.explore(m)
m = scene.intersection(request).explore(m)
m
```

<img width="507" height="495" alt="Image" src="https://github.com/user-attachments/assets/29c836fd-f110-4842-a731-071bcdf3bc54" />

<img width="548" height="489" alt="Image" src="https://github.com/user-attachments/assets/7f273462-b264-4f22-a317-83cf9369128b" />

PR #241, opened 2025-08-06
fix: exposing simplify from Geometry.geojson by kwargs in explore

MarkusZehner:
Fix for issue #240 

Simple workaround to expose simplify to the explore function.

### DIFF
--- a/odc/geo/geom.py
+++ b/odc/geo/geom.py
@@ -28,7 +28,7 @@ from typing import (
 import numpy
 from affine import Affine
 from pyproj.aoi import AreaOfInterest
-from shapely import geometry, ops
+from shapely import geometry, ops, count_coordinates
 from shapely.coords import CoordinateSequence
 from shapely.geometry import base
 
@@ -849,6 +849,7 @@ class Geometry(SupportsCoords[float]):
         self,
         map: Optional[Any] = None,
         tiles: Any = "OpenStreetMap",
+        simplify: float | Literal["auto"] = "auto",
         attr: Optional[str] = None,
         map_kwds: Optional[Dict[str, Any]] = None,
         **kwargs: Any,
@@ -864,6 +865,10 @@ class Geometry(SupportsCoords[float]):
             Map tileset to use for the map basemap. Supports any option
             supported by :py:mod:`folium`, including "OpenStreetMap",
             "CartoDB positron", "CartoDB dark_matter" or a custom XYZ URL.
+        :param simplify:
+            Optionally simplify geometries by tolerance in degrees.
+            By default if fewer than 100 points in geometry 0.0,
+            else min(0.05, res*0.1).
         :param attr:
             Map tile attribution; only required if passing custom tile URL.
         :param map_kwds:
@@ -884,9 +889,10 @@ class Geometry(SupportsCoords[float]):
 
         # Convert to GeoJSON with resolution based on approx 100
         # points per side for proper plotting/reprojection
-        bbox = self.boundingbox
-        res = max(bbox.span_x, bbox.span_y) / 100
-        geojson = self.geojson(resolution=res)
+        res = _auto_resolution(self)
+        if simplify == "auto":
+            simplify = _auto_simplify(self)
+        geojson = self.geojson(resolution=res, simplify=simplify)
 
         # Create layer and add to map
         layer = GeoJson(data=geojson, **kwargs)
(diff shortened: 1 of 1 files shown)


Kirill888:
@MarkusZehner thanks for this. Can we make it more "finished" though. Ideally we should address #240 not just provide an undocumented work-around.

I think we need `simplify: float | Literal["auto"] = "auto"` argument exposed as an optional parameter, with it's own entry in the function signature and it's own documentation string.

The default `"auto"` case should resolve to `simplify=0` if geometry has "few enough" points (let's say 100, that should display quickly enough). Otherwise it should be `simplify=min(0.05, res*0.1)`. Can you implement that and check if this fixes your problem case?

Please ensure the following:

1. Run `black` on the file you have edited
2. Do not make a new commit, instead `git amend` your current commit with the new changes, then use `git push --force-with-lease` to submit your new work
3. Please make sure git commit message width fits within 80 char column and follows this pattern
    ```
    Summary of Changes #240 (<80 char)

    ^ empty line and more description when needed.
    Wrap text to make sure commit message fits within 80 char column limit.
    Reference issue you are fixing with #240 syntax, usually in the summary line
    ```

pinging @robbibt as the original author of `.explore`

MarkusZehner:
I am running into some problems calling Geometry.points 
```
---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)

File /python3.12/site-packages/odc/geo/geom.py:586, in Geometry.points(self)
    585 @property
--> [586](/python3.12/site-packages/odc/geo/geom.py:586) def points(self) -> CoordList: return self.coords

File /python3.12/site-packages/odc/geo/geom.py:584, in Geometry.coords(self)
    583 @property
--> [584](/python3.12/site-packages/odc/geo/geom.py:584) def coords(self) -> CoordList: return cast(list[tuple[float, float]], list(self.geom.coords))

File /python3.12/site-packages/shapely/geometry/polygon.py:266, in Polygon.coords(self)
    263 @property
    264 def coords(self):
    265     """Not implemented for polygons."""
--> [266](/python3.12/site-packages/shapely/geometry/polygon.py:266)     raise NotImplementedError(
    267         "Component rings have coordinate sequences, but the polygon does not"
    268     )

NotImplementedError: Component rings have coordinate sequences, but the polygon does not
```

However, `len(geometry.exterior.coords)` does give the number of coords. Should this be checked when calling Geometry.points or Geometry.coords?

Kirill888:
@MarkusZehner good point, I forgot it's not that easy to find "number of points" making up an arbitrary geometry object. Just counting exterior points of a polygon is not good enough for a polygon with fractal holes within it. This approach will also break for more complex geometries like feature collections.

I'm not aware of an obvious shapely method that can return number of nodes geometry is made up of. Every type has a different method for exposing point nodes it's made of, and requires marshaling all those points through C-Python interface just to reduce that back to a single number. We can probably use `.wkb` method to get a rough idea of geometry "weight", but for now, let's just skip that and always perform simplification when in "auto" mode, but make it more robust by selecting appropriate simplification threshold based on resolution used.

Kirill888:
Also, please rebase against the current `develop` branch to get rid of CI errors

MarkusZehner:
@Kirill888 how about using [shapely.count_coordinates](https://shapely.readthedocs.io/en/2.1.1/reference/shapely.count_coordinates.html), also the resolution is dependent on crs which hinders a simple threshold.

While setting up `_auto_simplify`, I stumbled upon `_auto_resolution`. By docs and comments, both aim for the same 100pt/side, while not resulting in the same value. For simplicity I would opt to use  `_auto_resolution` in explore.

odc/geo/geom.py
```diff
@@ -864,6 +865,10 @@ def explore(
             Map tileset to use for the map basemap. Supports any option
             supported by :py:mod:`folium`, including "OpenStreetMap",
             "CartoDB positron", "CartoDB dark_matter" or a custom XYZ URL.
+        :param simplify:
+            Optionally simplify geometries by tolerance in degrees.
+            By default if fewer than 100 points in geometry 0.0,
+            else min(0.05, res*0.1).
```
Kirill888, line 871: Too much detail, `res` is not an argument, we should not mention it, 100 is an implementation detail, also no need to mention. Mostly this will be used as `simplify=0`, so explain why would one use it.

odc/geo/geom.py
```diff
@@ -1475,3 +1481,9 @@ def mid_longitude(geom: Geometry) -> float:
 def _auto_resolution(g: Geometry) -> float:
     # aim for ~100 points per side of a square
     return math.sqrt(g.area) * 4 / 100
+
+
+def _auto_simplify(g: Geometry) -> float:
+    r = _auto_resolution(g.to_crs(4326))
+    s = 0.0 if count_coordinates(g.geom) <= 100 else min(0.05, r * 0.1)
+    return s
```
Kirill888, line 1489: this doesn't feel right:

- converting to lonlat twice
- converting even when not needed (i.e. `s=0.0` case)
- function not generic enough to be at module level, or be a function at all

Maybe it makes more sense to move `simplify="auto"` processing into `.geojson` method

Kirill888:
See #248 instead

PR #242, opened 2025-08-09
CI: use OIDC for CodeCov

pjonsson:
This uses a short-lived token
which is better for security.

Also pin the actions by hash
to make it harder to get
compromised if some action
gets compromised.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,10 +9,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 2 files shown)


pjonsson:
Considering the code coverage works, I don't think the CI failures are caused by this PR.

omad (approved):

omad:
> Considering the code coverage works, I don't think the CI failures are caused by this PR.

Agreed, thanks Peter, merging now.

PR #245, opened 2025-08-12
Fix CI/CD

Kirill888:
* mypy
* docs build

### DIFF
--- a/odc/geo/_dask.py
+++ b/odc/geo/_dask.py
@@ -139,6 +139,6 @@ def dask_rio_reproject(
             b_shape = tuple(ch[i] for ch, i in zip(dst_chunks, idx))
             dsk[k] = (np.full, b_shape, fill_value, src.dtype)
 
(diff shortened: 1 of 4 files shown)


odc/geo/_dask.py
```diff
@@ -139,6 +139,6 @@ def _src(idx: dict):
             b_shape = tuple(ch[i] for ch, i in zip(dst_chunks, idx))
             dsk[k] = (np.full, b_shape, fill_value, src.dtype)
 
-    dsk = HighLevelGraph.from_collections(name, dsk, dependencies=(src,))
+    dsk = HighLevelGraph.from_collections(name, dsk, dependencies=(src,))  # type: ignore
```
pjonsson, line 142: With MyPy, it's possible to do `type: ignore[attr-defined]` to only ignore the attr-defined and still get reports of any other type errors on the line.

pjonsson (approved):
Not having a broken CI is pretty good, merging this right away.

PR #246, opened 2025-08-12
fix: don't paste if axis is flipped #244

### LINKED ISSUE #244: `compute_reproject_roi` broken for "flippped" axis case
`odc.geo.overlap.compute_reproject_roi` is a bit of an orphan, it's not used in `odc-geo`. It is however an integral part of the `odc-loader`, and hence the effects are visible to users of `odc-stac`. One can think of this as a sort of "load planner". Given destination pixel grid and the source image pixel grid, this method computes what part of the source image should be loaded and whether one can just paste section of the source image into the section of the destination image.

Originally reported in `odc-stac` repo:

* https://github.com/opendatacube/odc-stac/issues/230
* (possibly related) https://github.com/opendatacube/odc-stac/issues/217

When source and destination images are related via "axis flip" (mirror operation), this method incorrectly reports that "pasting source pixels to the destination grid is OK", but the corresponding "index into source" image lacks flip (i.e. `step=-1` component of the python slice object). The outcome is that one ends up with an upside down image when "y axis is flipped", mirror image is returned when "x axis is flipped".

I think best approach here is to not even attempt to correctly compute source and destination indexes when axis direction is reversed between the two pixel grids, instead we should return `.paste_ok=False`. This will force loading code to take "normal" *reproject* route instead of *paste* route in this scenario. Given that this situation is rare, any performance degradation or image quality loss that can result from avoiding direct paste are not as important.

Kirill888:
Don't consider pasting from source to destination
when computing reproject roi if axis is flipped.

Closes #244

### DIFF
--- a/odc/geo/overlap.py
+++ b/odc/geo/overlap.py
@@ -385,6 +385,9 @@ def _can_paste(
 
     (sx, _, tx, _, sy, ty, *_) = A_  # tx, ty are in dst pixel space
 
+    if any(s < 0 for s in (sx, sy)):
+        return False, "flipped axis"
+
     # Expect identity for scale change
     if any(abs(abs(s) - 1) > stol for s in (sx, sy)):  # not equal scaling across axis?
         return False, "sx!=sy, probably"
--- a/tests/test_overlap.py
+++ b/tests/test_overlap.py
@@ -427,20 +427,23 @@ def test_axis_overlap() -> None:
 def test_can_paste() -> None:
     assert _can_paste(mkA(translation=(10, -20))) == (True, None)
     assert _can_paste(mkA(scale=(10, 10))) == (True, None)
-    assert _can_paste(mkA(scale=(-10, 10), translation=(0, -4 * 10))) == (True, None)
+    assert _can_paste(mkA(scale=(-10, 10), translation=(0, -4 * 10))) == (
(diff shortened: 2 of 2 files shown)


pjonsson (approved):
I don't know the surrounding code, but taking the slow path is better than being wrong.

odc/geo/overlap.py
```diff
@@ -385,6 +385,9 @@ def _can_paste(
 
     (sx, _, tx, _, sy, ty, *_) = A_  # tx, ty are in dst pixel space
 
+    if any(s < 0 for s in (sx, sy)):
+        return False, "flipped axis"
+
     # Expect identity for scale change
     if any(abs(abs(s) - 1) > stol for s in (sx, sy)):  # not equal scaling across axis?
```
harryC-space-intelligence, line 392: One question on this line - not particularly important as this fixes my original issue but.

Is there a need for the two calls to `abs`?

I found that changing this line to `if any(abs(s - 1) > stol for s in (sx, sy)):` made more sense in my head and also meant that `can_paste` is set to false if any axes are flipped.

Kirill888: yes, especially since "reason" messages aren't used, most likely, probably an early debugging tool

PR #248, opened 2025-08-19
Smarter handling of geojson in explore

### LINKED ISSUE #240: default simplify value in odc.geo.geom.Geometry.geojson results in misleading display for small shapes

Kirill888:
* Automatic computation of `simplify` argument in `Geometry.geojson`
  * don't simplify small geometries (fewer than 100 nodes)
  * Use 1% of the largest span of the geometry in lonlat OR 0.05 degrees, whichever is smallest
* Pass through all `.geojson` parameters from `.explore` (`resolution`, `wrapdateline` and `simplify`) 
* expose `shapely.count_coordinates`

Closes #240

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -142,6 +142,7 @@ Shapely geometry classes with CRS information attached.
    chop_along_antimeridian
    clip_lon180
    common_crs
+   count_coordinates
    densify
    intersects
    lonlat_bounds
--- a/odc/geo/geom.py
+++ b/odc/geo/geom.py
@@ -883,9 +883,13 @@ class Geometry(SupportsCoords[float]):
         :param map_kwds:
             Additional keyword arguments to pass to :py:class:`folium.Map`.
         :param kwargs:
-            Additional keyword arguments to pass to :py:class:`folium.GeoJson`.
+            Additional keyword arguments to pass to :py:meth:`odc.geo.geom.Geometry.geojson`
+            and :py:class:`folium.GeoJson`.
 
         :return: A :py:mod:`folium` map containing the plotted Geometry.
+
+        :seealso: :py:meth:`odc.geo.geom.Geometry.geojson` for more details on
+            the parameters.
         """
         # pylint: disable=import-outside-toplevel, redefined-builtin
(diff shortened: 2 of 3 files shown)


Kirill888:
@MarkusZehner it was easier to handle "auto" in `.geojson` method instead, this should also allow overriding `resolution` and supplying `wrapdateline` in explore.

PR #266, opened 2026-05-14
Use Ruff instead of black+isort

pjonsson:
Replace black and isort with Ruff,
and reformat to make Ruff happy.

I'm personally not a fan of pre-commit,
but since that is already used for other
formatting we put the Ruff modules in
there to get the CI to do formatting checks.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -89,33 +89,6 @@ jobs:
         run: |
           mamba env export -p /tmp/test_env
 
-  run-black-check:
-    timeout-minutes: 15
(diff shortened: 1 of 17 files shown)


pjonsson:
The pre-commit update in #267 is failing the same way, so the reformatting changes from this PR is not the cause of the test failure.

omad (approved):

pjonsson:
Actions are stuck stating they are waiting for a hosted runner to come online after I rebased. Merging this since it passed most of the CI yesterday.

PR #268, opened 2026-05-14
Update coords after ds reproject

andrewdhicks:
WIP - Dataset was retaining old coords when reprojected. `xr.map(reproject)` doesn't deal with coord changes to parent, and leaves the dataset->dataarray coords mismatched.

Now we build a new `Dataset` and guess what to copy over. I'm skipping known spatial attrs and coords on spatial dims. Seems flaky,


pjonsson:
@andrewdhicks I noticed your PRs tend to get reformatted by pre-commit. I think I've replaced black+isort everywhere now, so if you run `uv run ruff check --fix` and `uv run ruff format` before you commit in most ODC repositories, that should format the code and avoid pre-commit stepping in later and messing up git blame.

If you want to run ruff on this PR, I think you have to rebase to latest develop.

pjonsson:
@andrewdhicks I've been digging through git history in other ODC repositories and it's hard to know where commits came from and which earlier change that caused them to be added.

There are 6 consecutive commits that relax the test comparisons in this PR, and none of the commit messages of those commits contain any further information besides the title. I like being able to use `git bisect` and I assume these 6 commits breaks that. Is there any reason to keep them as separate commits, or can you squash the respective test relaxations into the commit that makes the code adjustments that causes the adjusted test to start failing?

odc/geo/_xr_interop.py
```diff
@@ -12,13 +12,15 @@
 import json
 import math
 import warnings
+from collections.abc import (
+    Hashable,
+)
 from dataclasses import dataclass
```
SpacemanPaul, line 16: This can be a single line.

odc/geo/crs.py
```diff
@@ -3,12 +3,14 @@
 # Copyright (c) 2015-2020 ODC Contributors
 # SPDX-License-Identifier: Apache-2.0
 import warnings
+from collections.abc import (
+    Hashable,
+)
 from typing import (
```
SpacemanPaul, line 7: One line.

SpacemanPaul (approved):
- Put `from collections.abc import Hashable` on one line
- Cleanup git history.

Otherwise good to merge.  Thanks Andrew.

andrewdhicks:
The failing tests were due to a stale cache for some pipeline steps.
I will address these issues in a separate PR.

tests/conftest.py
```diff
@@ -59,5 +59,5 @@ def country_raster(country, resolution):
 def country_raster_f32(country, resolution):
     geobox = GeoBox.from_geopolygon(country, resolution=resolution, tight=True)
     xx = rasterize(country, geobox)
-    xx = xr.where(xx, np.random.uniform(0, 100, xx.shape).astype("float32"), 0)
```
Kirill888, line 62: this change swaps mask and value arrays see #290

PR #272, opened 2026-06-03
Update pipeline to match other repos

andrewdhicks:
The current CI pipeline is using a cached conda environment which is not always installing the most recent files in the test steps.
This change switches to using `uv` in line with other ODC repos such as `odc-loader`.

This also removes the docs preview and publishing steps from the CI pipeline. The recommended integration with readthedocs is to use webhooks. Actions are no longer supported.


.github/workflows/release.yml
```diff
@@ -0,0 +1,54 @@
+---
+name: Build & upload PyPI package
```
pjonsson: If you take the last 15-20 lines of `main.yml` from datacube-core and put those lines at the end of `main.yml` in this repository, we'll get the package checked in every PR so we can't mess up the metadata before the release, and the upload to PyPi will only engage after the tests and package checks have been successful.

(Since main in this repository is using a test matrix you probably need a separate job gated on completion of the test job for the package check/PyPi upload.)

.github/actions/setup-python-env/action.yml
```diff
@@ -0,0 +1,32 @@
...
+
+inputs:
+  python-version:
+    description: "Python version to use"
+    required: true
+    default: "3.12"
+  uv-version:
+    description: "uv version to use"
+    required: true
+    default: "0.5.6"
```
pjonsson: This version is ancient, the current release is 0.11.17.

odc/geo/_blocks.py
```diff
@@ -126,9 +126,7 @@ def extract(
         *,
         dtype=None,
         roi=None,
-        casting: (
-            Literal["no", "equiv", "safe", "same_kind", "unsafe"] | None
-        ) = "same_kind",
+        casting: (Literal["no", "equiv", "safe", "same_kind", "unsafe"]) = "same_kind",
```
pjonsson: The numpy documentation does say this parameter is optional, are you sure this is the right type annotation?

If the `| None` can be removed, the parentheses can also be removed since it fits in a single line.

tests/test_xr_interop.py
```diff
@@ -634,9 +634,11 @@ def test_reload(geobox) -> None:
     assert ds.odc.reload() is ds
     assert ds.odc.geobox == geobox
 
-    # change coords in-place
-    coord = xx[list(xx.coords)[0]]
-    coord.data[:] += 0.337
+    # change coords - copy as can be read-only
+    coord_name = list(xx.coords)[0]
```
pjonsson, line 638: The previous code was added to test in-place changes, so I don't think this is what we want.

odc/geo/_blocks.py
andrewdhicks: It's optional, but the default value is the string literal `"same_kind"`, so mypy was raising `None` as an error.
I'll remove the parentheses.

tests/test_xr_interop.py
andrewdhicks: Newer versions of xarray enforce immutability of the data values so it stays in sync with the pandas index on the coordinate. This tests that the coordinate can be changed and that the geobox can be recalculated with `reload()`.

.github/workflows/release.yml
andrewdhicks: That's a good idea. I'll sort out the matrix gate when I am back next week.
The `hynek/build-and-inspect-python-package` action includes the twine check, so it could be on the PR trigger in main, with only the upload happening on the tag trigger?

andrewdhicks: I've merged the release steps into main, and gated build to happen after the tests.
Also only uploading the coverage from 3.12 now.

.github/workflows/main.yml
```diff
@@ -3,6 +3,10 @@ name: Run Code Checks
 on:
   pull_request:
   push:
+    tags:
+      - "v*"
```
pjonsson: Does this really work in the way you intended? My guess would be that the new behavior is either that the CI only runs when there are tags created (so not for PR merges to main), and/or that we get two workflows when making a release.

.github/actions/setup-python-env/action.yml
```diff
@@ -0,0 +1,32 @@
...
+  steps:
+    - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+      with:
+        python-version: ${{ inputs.python-version }}
+
+    - name: Install uv
+      uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
+      with:
+        version: ${{ inputs.uv-version }}
+        enable-cache: "true"
```
pjonsson: This cache should be disabled when publishing releases, otherwise we're susceptible to some kind of cache poisoning attack.

(It's an issue in other repositories as well, but I have too many things on my plate so I haven't gotten around to fixing it yet.)

.github/workflows/main.yml
```diff
@@ -11,302 +16,98 @@ concurrency:
...
-          echo "/tmp/test_env/bin" >> $GITHUB_PATH
-
-      - name: Install in Edit mode
-        shell: bash
-        run: |
-          pip install -e . --no-deps
-
-      - name: Check with pylint
+          echo "🚀 Checking lock file consistency with 'pyproject.toml'"
+          uv lock --locked
```
pjonsson: I think this consistency check is ensured by running pre-commit in the next line.

.github/workflows/main.yml
```diff
@@ -11,302 +16,98 @@ concurrency:
...
+      - name: Run tests
+        run: uv run --python ${{ matrix.python-version }} pytest --cov=odc --cov-report=xml --durations=5 tests
 
-      - name: Upload Coverage
-        if: |
-          github.repository == 'opendatacube/odc-geo'
-
-        uses: codecov/codecov-action@18283e04ce6e62d37312384ff67231eb8fd56d24 # v5.4.3
+      - name: Upload coverage to Codecov
+        if: matrix.python-version == '3.12'
```
pjonsson: Doesn't CodeCov merge coverage files behind the scenes, so uploading in every job gives a better indication of what is actually tested?

.github/workflows/main.yml
```diff
@@ -11,302 +16,98 @@ concurrency:
...
-      - build-test-env-base
-
+      id-token: write  # Permit authenticating to CodeCov.
+    defaults:
+      run:
+        shell: bash
     steps:
-      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Check out
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
pjonsson, line 57: The switch to setuptools-scm is in a different PR, but I suggest setting `fetch-depth: 0` for the checkouts that are used to run code/build packages, so they get the right version number.

.github/workflows/main.yml
```diff
@@ -11,302 +16,98 @@ concurrency:
...
-        shell: bash
-        run: |
-          echo "/tmp/test_env/bin" >> $GITHUB_PATH
-
-      - name: Check with mypy
-        shell: bash
-        run: |
-          mypy --explicit-package-bases --python-version 3.10 odc
+        env:
+          DASK_TEMPORARY_DIRECTORY: /tmp/dask
```
pjonsson: Is this variable needed for running pre-commit or MyPy?

.github/workflows/main.yml
```diff
@@ -11,302 +16,98 @@ concurrency:
...
-          --cov-report=html \
-          --cov-report=term \
-          --cov-report=xml:coverage.xml \
-          --timeout=30 \
-          tests
-
-        env:
-          DASK_TEMPORARY_DIRECTORY: /tmp/dask
+      - name: Run tests
+        run: uv run --python ${{ matrix.python-version }} pytest --cov=odc --cov-report=xml --durations=5 tests
```
pjonsson, line 69: It's a bit hard to get an overview in the GitHub diff view, is this meant to run without `DASK_TEMPORARY_DIRECTORY`?

.github/workflows/main.yml
```diff
@@ -11,302 +16,98 @@ concurrency:
...
 
-      - name: Update PATH
-        shell: bash
-        run: |
-          echo "/tmp/test_env/bin" >> $GITHUB_PATH
+  release-pypi:
+      name: Publish released package to pypi.org
+      timeout-minutes: 15
+      environment: release-pypi
+      if: github.repository_owner == 'opendatacube' && github.event_name == 'release' && github.event.action == 'published' && startsWith(github.event.release.tag_name, 'v')
```
pjonsson: I'm not sure what behavior is desired, but I know that datacube-core doesn't have a check on the tag name for uploading a release. What is the reasoning for "ignoring" published releases with the "wrong" format of the tag?

.github/workflows/main.yml
```diff
@@ -11,302 +16,98 @@ concurrency:
...
+      - name: Check out
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
 
-      - uses: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809 # v4.2.4
-        id: conda_cache
-        with:
-          path: /tmp/test_env
-          key: ${{ runner.os }}-test-env-py310-${{ hashFiles('tests/test-env-py310.yml') }}
```
pjonsson, line 60: Should the Conda test env files be removed now that the CI stops using them?

.github/actions/setup-python-env/action.yml
pjonsson: I don't know anything about reusable workflows, but here's the syntax for the `setup-uv` action in a regular workflow according to zizmor: `enable-cache: ${{ !startsWith(github.ref, 'refs/tags/') }}`.

pjonsson:
This PR has been hanging around for 3 weeks already and it keeps growing, could you please split out the code fixes to separate PRs so we can get them merged so the CI isn't constantly red? (And a lot of these fixes can be reviewed in 1 minute and then we would never have to think about them again, so we might even save effort on that.)

.github/workflows/main.yml
```diff
@@ -11,302 +14,97 @@ concurrency:
...
     timeout-minutes: 15
     runs-on: ubuntu-latest
-
     steps:
-      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Check out
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
+          fetch-depth: 0
```
pjonsson, line 25: This change was in preparation for setuptools-scm, so I guess that can be dropped now.

.github/workflows/main.yml
```diff
@@ -11,302 +14,97 @@ concurrency:
...
+  release-pypi:
+      name: Publish released package to pypi.org
+      timeout-minutes: 15
+      environment: release-pypi
+      if: github.repository_owner == 'opendatacube' && github.event_name == 'release' && github.event.action == 'published'
+      runs-on: ubuntu-latest
+      needs: build-package
+
+      permissions:
+        id-token: write
```
pjonsson, line 100: This job is small right now, but it's still good to have a comment about what step that requires the token.

odc/geo/overlap.py
```diff
@@ -184,7 +184,7 @@ class ReprojectInfo:
     scale: float
     """Scale change as a single number. (the min of scale2)"""
 
-    scale2: XY[float]
+    scale2: XY[float] | XY[int]
```
pjonsson, line 187: In most other code we put the `|` as deeply nested as possible in type signatures, is there any difference between this signature and `XY[int | float]`?

tests/test_gcp.py
```diff
@@ -32,6 +32,10 @@ def test_gcp_geobox_from_rio(au_gcp_rio) -> None:
     assert src.height == gbox.height
 
 
+@pytest.mark.xfail(
+    True,
+    reason="Reprojecting gcp geobox is giving different results between python versions.",
```
pjonsson, line 37: We have 4 Python versions, can't `assert x == y` be rewritten to `assert x in [y1, y2]` or `assert (x == y1) or (x == y2)` to keep the test active?

Edit: and if that doesn't work, can't the test be marked as `skipif()` that compares the Python version, so at least the CI runs the test?

tests/test_gcp.py
```diff
@@ -92,17 +96,17 @@ def test_gcp_geobox_basics(au_gcp_geobox: GCPGeoBox) -> None:
...
     # map bounds without CRS, should still work, since data is in epsg:4326
     _mapping = GCPMapping(gbox._mapping._pix, gbox._mapping._wld, None)
     assert _mapping.crs is None
     gbox_ = GCPGeoBox(gbox.shape, _mapping)
     assert gbox_.crs is None
-    p3, p4 = gbox_.map_bounds()
-    assert p3 == pytest.approx(p1, abs=1e-3)
-    assert p4 == pytest.approx(p2, abs=1e-3)
+    p1, p2 = gbox_.map_bounds()
+    assert p1 == pytest.approx((-44.50301336231415, 109.39806656168265))
```
pjonsson, line 108: I get that the numbers in the result might change between Python versions (different GDAL/PROJ versions, different Python libraries, and rounding on top of that), but why can't these points be compared to the previous points like the test was before?

And even if there's a good reason to compare against the raw numbers here, I think it's nice to keep the previous point names instead of redefining existing points.

SpacemanPaul (approved):
We need to get this merged.  The issue around the inconsistent tests can be revisited in a future PR.

There are some other minor comments from PJ that could be addressed here or left for future PRs.

Please apply as a squash merge as the internal PR history is a mess.

tests/test_gcp.py
andrewdhicks: Ideally this would test for correct values, not values that just pass the test.
It is unclear what the correct values are.

odc/geo/overlap.py
andrewdhicks: See https://github.com/opendatacube/odc-geo/pull/294

.github/workflows/main.yml
andrewdhicks: See https://github.com/opendatacube/odc-geo/pull/294

.github/workflows/main.yml
andrewdhicks: See https://github.com/opendatacube/odc-geo/pull/294

.github/workflows/main.yml
andrewdhicks: See https://github.com/opendatacube/odc-geo/pull/294

tests/test_gcp.py
Kirill888: I feel like switching to `pypi` based testing might be contributing to the issues being observed. We used to test against conda based setup, that ensures that only 1 version of common libraries like `geos` and `proj`  are present on the system. With `uv` we rely on packages from `pypi` that can not, by design, share common libraries, at best you end up with identical copies of shared libraries coming from `rasterio`, `osgeo`, `pyproj` and `shapely`. Depending on import order you end up with one or the other `libgeos` or `libproj` being loaded at any given time, who knows what that breaks. Given that errors are intermittent strongly suggests that issue might be due to those runtime differences in code loading into the process from third party copiled libraries.

@andrewdhicks @pjonsson

Kirill888: but also tests can can be written in a more robust way for sure, but writing high quality tests for numeric processing is tricky, especially when testing an effectively "black box", like coordinate transformation step.

PR #274, opened 2026-06-09
Use SCM for versioning

andrewdhicks:
Follows the pattern of other repos, such as `odc-loader` to use git tags for releases.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "odc-geo"
 description = "Geometry Classes and Operations (opendatacube)"
-version = "0.5.1"
+dynamic = ["version"]
 authors = [
     {name = "Open Data Cube"},
 ]
@@ -50,11 +50,16 @@ las = ["laspy", "lazrs"]
 all = ["rasterio", "tifffile", "imagecodecs", "dask[array,distributed]", "boto3", "azure-storage-blob", "laspy", "lazrs"]
 
 [build-system]
-requires = ["flit_core >=3.2,<4"]
-build-backend = "flit_core.buildapi"
+requires = ["setuptools>=69", "setuptools_scm[toml]>=6.2"]
+build-backend = "setuptools.build_meta"
 
-[tool.flit.module]
-name = "odc.geo"
+[tool.setuptools.packages.find]
+exclude = ["tests"]
+include = ["odc*"]
+
+[tool.setuptools_scm]
+write_to = "odc/geo/_version.py"
+fallback_version = "0.5.2.dev0"
 
 [tool.mypy]
 python_version = "3.10"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ las = ["laspy", "lazrs"]
 all = ["rasterio", "tifffile", "imagecodecs", "dask[array,distributed]", "boto3", "azure-storage-blob", "laspy", "lazrs"]
 
 [build-system]
-requires = ["setuptools>=69", "setuptools_scm[toml]>=6.2"]
+requires = ["setuptools>=69", "setuptools_scm[toml]>=8,<10"]
 build-backend = "setuptools.build_meta"
 
 [tool.setuptools.packages.find]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,11 +8,10 @@ authors = [
 maintainers = [
     {name = "Open Data Cube"},
 ]
-license = {text = "Apache License 2.0"}
+license = "Apache-2.0"
 readme = "README.rst"
 keywords = ["gis", "geospatial", "opendatacube"]
 classifiers = [
-    "License :: OSI Approved :: Apache Software License",
     "Intended Audience :: Developers",
     "Operating System :: OS Independent",
     "Programming Language :: Python :: 3.10",
--- a/.gitignore
+++ b/.gitignore
@@ -150,3 +150,4 @@ cython_debug/
 docs/_api/
 tt.py
 tt_*py
+odc/geo/_version.py
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,2 @@
+prune docs
+prune tests
--- a/uv.lock
+++ b/uv.lock
@@ -1902,7 +1902,6 @@ wheels = [
 
 [[package]]
 name = "odc-geo"
-version = "0.5.1"
 source = { editable = "." }
 dependencies = [
     { name = "affine" },


pyproject.toml
```diff
@@ -50,11 +50,16 @@ las = ["laspy", "lazrs"]
 all = ["rasterio", "tifffile", "imagecodecs", "dask[array,distributed]", "boto3", "azure-storage-blob", "laspy", "lazrs"]
 
 [build-system]
-requires = ["flit_core >=3.2,<4"]
-build-backend = "flit_core.buildapi"
+requires = ["setuptools>=69", "setuptools_scm[toml]>=6.2"]
```
pjonsson: I'm fairly sure setuptools-scm 6.x isn't supported by something we care about because I remember bumping that lower bound in some other ODC repository around 2 (?) years ago. And it wasn't that many months since setuptools-scm 10.x broke something in core, but I've still forgotten what it was.

I realize my descriptions are not very helpful, but I suggest you make the setuptools-scm version constraint `>=8,<10` for now so we can get this merged.

pyproject.toml
```diff
@@ -50,11 +50,16 @@ las = ["laspy", "lazrs"]
...
+build-backend = "setuptools.build_meta"
 
-[tool.flit.module]
-name = "odc.geo"
+[tool.setuptools.packages.find]
+exclude = ["tests"]
```
pjonsson, line 56: If you install the Python `build` package and run `python3 -m build .` on the main branch and then `mv dist dist.orig` and then run the same build on this branch, the source archive on this branch contains a lot more files.

The `setup.cfg` that wasn't there before is expected, but all the configuration/CI/docs should probably be omitted in the future as well. I don't know the mechanics for this stuff, but the reason "file selection" works for core is because there's a `MANIFEST.in` detailing what should be packaged.

Unless you know how to get this to behave the way we want, I suggest just adding a  `MANIFEST.in` for now.

pjonsson:
The changes that are here more or less looks like I would expect them to look, it's the missing `MANIFEST.in` (or equivalent functionality in some other way) that is the main problem.

Edit: and when I build the package, I'm getting a bunch of deprecation warnings about the `pyproject.toml` format, could you please make a separate PR which fixes those things while you're looking into this?

pyproject.toml
andrewdhicks: The docs and tests are getting pulled in by setuptools_scm, so the normal setuptools.packages.find include/exclude properties were being ignored. Adding the manifest fixed it. Thanks!

Kirill888:
can we not? I personally found no benefit in using SCM based versioning. It causes a lot of pains, and the gains are rather slim.

Kirill888:
BTW `odc-{loader,stac,geo}` all declare version manually inside `pyproject.toml`, then use importlib to report version in code. Dynamic version adds a lot of complexity to the distribution pipeline: need to checkout enough depth, constantly changing `uv.lock`, `pixi.lock`. And `setuptools` itself is just not great. This will also require changes to `conda-forge` publishing too.

I feel that explicit version bump as the last tagged commit before release is a more obvious, more explicit way of managing release versions.

SpacemanPaul:
Consistency across the project would be nice, but is unachievable given the odc-tools packages.

IMO if the release process needs anything beyond clicking "New Release" in github (i.e. bumping the version number in pyproject.toml in PT), then that needs to be documented, either in the README, or in the dev docs if they exist.

Therefore we should either:

a) switch odc-geo (and odc-loader and odc-stac) to SCM versioning; OR
b) document in those projects' READMEs that the version number needs to manually bumped in a PR prior to the release.

Kirill888:
> a) switch odc-geo (and odc-loader and odc-stac) to SCM versioning; OR
b) document in those projects' READMEs that the version number needs to manually bumped in a PR prior to the release.

I agree that release process needs to be documented and possibly automated more. I don't see how switching to scm based versioning makes release process any simpler or indeed remove the need for the documentation. If anything scm based versioning has more indirection and requires a larger document to cover the gotchas it introduces. The things I dislike the most about it is the need to create unversioned files that DO need to be shipped as part of the source distribution. It also makes it harder to see where releases happen as you looks through the git history as there are no commits with version bumps, only tags. To me this hides important information in the "metadata" instead of explicitly capturing this in the "data", which is versioned code in this case. That does not mean, that those code changes need to be done manually by a human, an action an authenticated user can trigger on a specific commit can very well bump the version, commit the change, add the tag and push all of that to `main` triggering publish to pypi if all goes well.

@SpacemanPaul @andrewdhicks in your opinion which of the active `opendatacube/` repos have the most reasonable automated release process that should be adopted by all the other? Which `release.yaml` should we aim to replicate across as many repos as possible? And does that require SCM based versioning?

andrewdhicks:
My main aim is consistency across the repos, I don't have strong opinions on the actual method.
I've been following the pattern from `odc-loader`, mostly for the other CI steps.

Kirill888:
thanks @andrewdhicks , `odc-loader` does have automated release from github, but it does not use SCM for version management, latest PR that bumped version manually is here: https://github.com/opendatacube/odc-loader/pull/44

SCM was originally attempted by Damien in that repo, but he then switched to hard-coded versions early on in the project existence: https://github.com/opendatacube/odc-loader/commit/a1fddbff9eb820224aecb7d19d99ea7c37060199 I believe due to issues with `uv.lock`

pjonsson:
> And `setuptools` itself is just not great.

I don't think anyone considers setuptools great, I'm guessing its dominance is because it used to be more or less the only choice. There are alternatives to it nowadays though, and since we're already using uv in a lot of repositories, it's possible the [uv build backend](https://docs.astral.sh/uv/concepts/build-backend/) should be considered for repositories not using VCS based versioning.

I don't do a lot of releases, but CI-wise, I think datacube-core has the simplest structure for the CI among the ODC repositories. There's a (uv-based) main and lint workflow that does the majority of the stuff and mainly runs on code changes, and then there's a doc-preview for special permissions for ReadTheDocs, a doc-qa that runs on everything, and a scan workflow that does nightly runs of Trivy for the security tab. Uploading to PyPi is gated on tests passing in the main workflow, so we might upload packages with bad formatting to PyPi, but the code will have passed the tests successfully.

Kirill888:
@pjonsson datacube uses docker-compose for running tests, understandable as it requires database setup, but hardly the simplest setup. Dynamic version from SCM is also supported by other build backends, but issues with that approach still hold, regardless of the backend.

What I would like to see as an optimal release pipeline:

1. Runs in response to `Release.published`
2. Checks that running off a clean checkout off a tag that matches the version of the package
3. Builds wheel and source distributions for upload to pypi
4. Runs tests against an environment built from the wheel to be published
    * ideally no checkout, only artifact fetch for the wheel, all deps should come from the package, not `uv.lock`
    * `pip install ./some-package.whl[tests]` into a clean env kinda thing
    * repeat for source distribution also
5. Push to pypi only if all steps above succeeded

pjonsson:
>Dynamic version from SCM is also supported by other build backends, but issues with that approach still hold, regardless of the backend.

I'm not disputing your technical arguments around the costs of using SCM versioning, they all seem sensible to me. I've used hatch+hatch-vcs internally, and while the technical details are different for those, the user experience is more or less the same as for setuptools-scm in my experience.

> 4. Runs tests against an environment built from the wheel to be published
>    
>    * ideally no checkout, only artifact fetch for the wheel, all deps should come from the package, not `uv.lock`
>    * `pip install ./some-package.whl[tests]` into a clean env kinda thing
>    * repeat for source distribution also

Personally I don't appreciate starting a bunch of different jobs that shuffle artifacts between them, it's faster to let the job that builds the artifacts just delete the repository if that level of isolation is required, and that also keeps everything for reproducing a CI run locally in a dozen consecutive lines in the yaml files (which is especially important for casual and drive-by contributors).

I think there are good technical arguments for both keeping locked dependencies in testing (reproducible errors, CI not suddenly going red without any code changes) and testing the user experience for someone who installs the package from PyPi (software without users isn't very useful). Whichever of those two choices one makes, the release workflow shouldn't fail if the pull request that bumped the version number was green 5 minutes earlier.

Kirill888:
> Personally I don't appreciate starting a bunch of different jobs that shuffle artifacts between them, it's faster to let the job that builds the artifacts just delete the repository if that level of isolation is required, and that also keeps everything for reproducing a CI run locally in a dozen consecutive lines in the yaml files (which is especially important for casual and drive-by contributors).
> 
> I think there are good technical arguments for both keeping locked dependencies in testing (reproducible errors, CI not suddenly going red without any code changes) and testing the user experience for someone who installs the package from PyPi (software without users isn't very useful). Whichever of those two choices one makes, the release workflow shouldn't fail if the pull request that bumped the version number was green 5 minutes earlier.

@pjonsson I was talking specifically about testing RELEASE packages. I don't care if it's done in separate workflows, separate jobs, separate steps or within the single step of a job. But unless you have tested the code installed from the package you have not tested packaging. You have tested that same code that was green when PR was merged, and assuming your tests are well-written, they should keep passing from that commit for the foreseeable future. BUT that doesn't test PACKAGING, you can still have a perfectly compliant package with all the required metadata that will be accepted by pypi without a single warning, but that doesn't ship all the code or data needed for the package to even import, yet alone run correctly. And of course you can also easily omit declaring an essential dependency that just happens to be present in you test harness environment.

To have some certainty that what is about to be pushed to pypi is "fit for purpose", one must install it into an empty python environment and then at least import all the sub-modules, but ideally also run tests against code installed from the wheel. Because running tests in situ doesn't catch packaging errors, like omission of data files.

PR #275, opened 2026-06-12
Accept Path destinations in save_cog_with_dask

### LINKED ISSUE #239: [ERROR] Error when specifying a Path object as destination for a COG
Currently running `odc-geo == 0.5.0rc1`.

I'm passing a `Path` object as destination to store a COG and it's raising an error in this line: https://github.com/opendatacube/odc-geo/blob/9804d5bb21aae6f58ab35c2b0efedc5e9667adde/odc/geo/cog/_tifffile.py#L780

I believe it would be useful and definitely _pythonic_ to support this input type, but I'm not sure how to proceed to PR it myself.

It is reproducible just by:

```python
from pathlib import Path
from odc.geo.cog import save_cog_with_dask

dst = Path("test.tiff")
...
d = save_cog_with_dask(xx, dst)
```

```bash
AttributeError: 'PosixPath' object has no attribute 'decode'
```

Maybe checking the type of the argument before calling `urlparse` could be a solution, but I'm wondering what implications can it have if we allow as well `S3Path`, `UPath` and other kinds of `Path` objects that extend python `Path`.

gaoflow:
Fixes #239.

`save_cog_with_dask(xx, Path("out.tif"))` crashed with `AttributeError: 'PosixPath' object has no attribute 'decode'` from `urlparse`. The destination is now normalized with `os.fspath` at function entry, so `Path` objects work the same as string paths — consistent with `write_cog`, which already accepts `Union[str, Path]`. Cloud URLs are unaffected (strings pass through `os.fspath` unchanged), and any other `os.PathLike` resolves to its filesystem path. Added a regression test that fails with the old AttributeError.

### DIFF
--- a/odc/geo/cog/_tifffile.py
+++ b/odc/geo/cog/_tifffile.py
@@ -10,8 +10,10 @@ from __future__ import annotations
 
 import itertools
 import math
+import os
 from functools import partial
 from io import BytesIO
+from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Optional, Union
 from urllib.parse import urlparse
 from xml.sax.saxutils import escape as xml_escape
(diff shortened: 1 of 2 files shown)


SpacemanPaul (approved):
LGTM

PR #286, opened 2026-06-20
readthedocs: build with uv

pjonsson:
Switch to using uv for the build,
and build on Ubuntu 24.04 using
Python 3.12.


pjonsson:
I don't see the PR being built by https://app.readthedocs.org/projects/odc-geo/builds/, so hard to tell if this works.

I have updated to a more recent geopandas though, commit d9410165c2c3 just stated it pinned the geopandas, but it looks like that was because of something that was later fixed in the tests (?).

omad (approved):
Sounds good, if it is broken we'll fix it.

PR #289, opened 2026-06-27
Fix doc build

pjonsson:
Change the documentation to refer
to where on the object instead
of passing the object as a parameter.
This makes xx.odc.geobox.crs
into non-None, so the assert on
native_crs not being None in
_map.py:135 passes.

Idea picked up from the conftest.py
change in #268.


SpacemanPaul (approved):

Kirill888:
@pjonsson @SpacemanPaul the change is not equivalent, original code used boolean array as mask and random data as value, so you end up with noise data limited to the footprint of a country. After the change it uses boolean array as value and random noise as mask.

PR #294, opened 2026-07-03
Apply suggestions from #272

andrewdhicks:


odc/geo/overlap.py
```diff
@@ -184,7 +184,7 @@ class ReprojectInfo:
     scale: float
     """Scale change as a single number. (the min of scale2)"""
 
-    scale2: XY[float] | XY[int]
+    scale2: XY[float | int]
```
Kirill888: I'm not comfortable with this change why is it needed? Feels like the fix should be at the calling sites not here,

I don't really think that mixing `float` and `int` under one structure is a good idea. Maybe going to `scale2: XY[float]` and then fixing calling sites would be more appropriate.

andrewdhicks: I've reverted it back to `XY[float]` and updated the one calling site that was causing the mypy error.

pjonsson (approved):

odc/geo/overlap.py
pjonsson: I agree, the callsite update does look better!

PR #297, opened 2026-07-06
Fix broken attr handling code when reprojecting datasets

andrewdhicks:


Kirill888 (approved):

odc/geo/_xr_interop.py
```diff
@@ -930,7 +930,11 @@ def _maybe_reproject(dv: xarray.DataArray):
 
     data_vars = {name: _maybe_reproject(da) for name, da in src.data_vars.items()}
     # Copy over non spatial-attrs and coords
-    attrs = {name: value for name, value in src.attrs if name not in SPATIAL_ATTRIBUTES}
+    attrs = {
+        name: value
+        for name, value in src.attrs.items()
```
Kirill888, line 935: I'm surprised no linter picked this up, and I guess no test had any attributes on a dataset either to trigger the problem.

Does this mean we need a new release? I feel like we do.

andrewdhicks: Yep, I think you are right.
I'm working on a python 3.12 version of `odc-stats`, and seeing what else shakes out with the new versions `odc-geo` and `odc-algo`. I've already found some MultiIndex issues with stats and the algo masking changes I made.

Issue #87, opened 2023-04-29
GeoboxTiles.tiles gives wrong results for global inputs


Kirill888:
`GeoboxTiles.tiles` takes in a geometry in any projection and returns an iterator over tiles that overlap with the given geometry. In order to do that, it first needs to project input shape into a CRS native to `GeoboxTiles`. This breaks down when source shape covers area way outside of the valid region of the destination CRS, as a result instead of returning all the tiles nothing is returned at all. Essentially this step `Geometry@crs1 -> BoundingBox(Geometry@crs2)` produces wrong results, this is especially common when source image has "global coverage" and destination CRS has a limited range like epsg:3577 for example.

What's needed is a robust "bounding box of a geometry in some other projection method", that does not slow down the common case where source shape is in a safe region for coordinate transformation. Ideally we would also have a "robust to_crs method for polygons" that would safely clip sections of the source geometry that are outside of the valid range of transform.

Berhinj:
ANy view if this is going to be fixed in the coming weeks/months?

PR #306, opened 2026-07-26
Don't lose tiles when projecting a large query into a limited-area CRS

binggao1230:
`GeoboxTiles.tiles()` returns **0 tiles** for a whole-world query against a 1680-tile EPSG:3577 grid, while the Australian subset of that same box returns 1512 — a superset query yielding less than its subset (#87).

Cause: the query's *vertices* are projected and the bounding box of the result is used as if it bounded the projected *region*, which only holds while the transform is near-affine over the query. The same line is in `range_from_bbox`, `GridSpec.tiles_from_geopolygon` and `GeoBox.from_geopolygon`, and it bites well short of "global" — a plain continental box drops 151 of 1663 tiles with no warning, and some UTM grids raise `Points of LinearRing do not form a closed linestring`.

Fix: densify before projecting, and clip the query to the target region when the destination CRS cannot represent it (lossless, since these callers only want the overlap with that region anyway). That needed the densification fixed first: `densify()` compared `x1**2 + x2**2` against `resolution**2` instead of the squared segment length, so a long segment near `x=0` was never split at any resolution, and a non-positive step never terminated — which `_auto_resolution()` returned for any zero-area geometry, so `to_crs(resolution="auto")` hung on a LineString.

Checked against a tile-by-tile oracle that projects each tile the other way (small, safe direction) and intersects in EPSG:4326, over 77 grid/query pairs across 7 CRSs: missing tiles 11087 → 0, spurious 3620 → 117, crashes 12 → 0. The remaining 117 are all on pole-covering grids, where a lon/lat box has no faithful image at all and over-selecting is the safe side; web mercator was already exact and stays exact. Full suite passes, and `densify()` no longer calls into shapely per point (~11x faster), which keeps timings within noise of before.

This is the narrower half of #87 — the existing call sites made robust, not the general-purpose robust `to_crs` for polygons you describe there. Happy to make `project_to_extent` private if you'd rather not commit to the name.

### DIFF
--- a/odc/geo/geom.py
+++ b/odc/geo/geom.py
@@ -458,10 +458,16 @@ def densify(
     """
     Adds points so they are at most `resolution` units apart.
     """
+    if len(coords) == 0:
+        return []
+    if not resolution > 0:
+        # nothing to add, and the loop below would never terminate
+        return [cast(Tuple[float, float], p) for p in coords]
+
     d2 = resolution**2
 
     def short_enough(p1, p2):
-        return (p1[0] ** 2 + p2[0] ** 2) < d2
+        return ((p1[0] - p2[0]) ** 2 + (p1[1] - p2[1]) ** 2) < d2
 
     new_coords: List[Tuple[float, float]] = [cast(Tuple[float, float], coords[0])]
     for p1, p2 in zip(coords[:-1], coords[1:]):
@@ -1532,4 +1538,10 @@ def count_coordinates(g: Geometry | Iterable[Geometry] | BoundingBox) -> int:
 
 def _auto_resolution(g: Geometry) -> float:
     # aim for ~100 points per side of a square
-    return math.sqrt(g.area) * 4 / 100
+    res = math.sqrt(g.area) * 4 / 100
+    if res > 0:
+        return res
+    # zero-area inputs (lines, degenerate polygons) still need densifying,
+    # fall back to the extent
+    bbox = g.boundingbox
+    return max(bbox.span_x, bbox.span_y) / 25
--- a/odc/geo/geom.py
+++ b/odc/geo/geom.py
@@ -1498,6 +1498,55 @@ def lonlat_bounds(
     return BoundingBox.from_xy(xx_range, bbox.range_y, crs=4326)
 
 
+def _projects_cleanly(geom: Geometry) -> bool:
+    """Did a projection produce a usable shape, or did it fall apart?"""
+    if geom.is_empty:
+        return False
+    return geom.is_valid and all(math.isfinite(v) for v in geom.boundingbox.bbox)
+
+
+def project_to_extent(geom: Geometry, extent: Geometry) -> Geometry:
+    """
+    Project ``geom`` into the CRS of ``extent``, robust to ``geom`` being large.
+
+    Transforming the vertices of a geometry and taking the result at face value only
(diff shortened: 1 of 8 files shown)


odc/geo/geom.py
```diff
@@ -1532,4 +1585,10 @@ def count_coordinates(g: Geometry | Iterable[Geometry] | BoundingBox) -> int:
 
 def _auto_resolution(g: Geometry) -> float:
     # aim for ~100 points per side of a square
-    return math.sqrt(g.area) * 4 / 100
+    res = math.sqrt(g.area) * 4 / 100
+    if res > 0:
```
pjonsson, line 1591: Can this check for `g.area > 0` and the previous return is inside the if-branch, or is this if-statement guarding against floating point precision issues?

If it is guarding against something subtle with precision, that should probably be described in a comment.

odc/geo/geom.py
```diff
@@ -458,24 +458,28 @@ def densify(
     """
     Adds points so they are at most `resolution` units apart.
     """
-    d2 = resolution**2
+    if len(coords) == 0:
+        return []
+    if not resolution > 0:
```
pjonsson, line 463: `if resolution <= 0.0`?

Kirill888: @pjonsson resolution can not be negative, but it might be zero, I think this is better handled with an assert to be honest

Kirill888:
@gaoflow thank you for the contribution, can I ask you to make a separate PR for the fix in densify? And then we can discuss other changes separately. Please split your changes into 3 commits

* fix for densify
* addition of new projection method
* fix for geotiles

Looks like `densify` was broken for six years by now, not too sure about proposed fix though, can we simply fix the error where it is which is:

https://github.com/opendatacube/odc-geo/blob/552bc58cb59519abaf7c7b920c48faac8e0fc76d/odc/geo/geom.py#L463-L464

should have been: 

```python
    def short_enough(p1, p2):
        return ((p1[0] - p2[0]) ** 2 + (p1[1] - p2[1]) ** 2) < d2
```

and not change the actual interpolation logic, do include your tests of course.

Kirill888:
looks like it's some kind of bot thing, it opened 300+ PRs that day, and ran out of tokens afterwards

binggao1230:
Done — the PR is now three commits: the densify fix, the new `project_to_extent`, and the geotiles fix.

I kept the densify change to the `short_enough` predicate (it now measures the actual segment) plus guards for empty / non-positive-`resolution` inputs that used to hang, with tests; the interpolation logic is untouched, as you suggested. The separate densify PR is being prepared on top of this.

And to be straight about the other comment: yes, these contributions are prepared with an AI coding agent, run under my account and responsibility.

binggao1230:
Separate densify PR is up: https://github.com/opendatacube/odc-geo/pull/310

odc/geo/geom.py
binggao1230: Not precision — area is never negative, this just tells apart a genuinely 2D geometry from a degenerate one (line/point/sliver) where sqrt(area) is a poor resolution estimate. Added a comment saying so in #310, since that's where this check ended up living after the split. Also worth noting: #310 additionally hardens `densify()` itself to tolerate `resolution<=0` (a single-point/empty geometry can still reach it via this fallback), so whichever ordering this check takes here no longer risks a crash either way.

PR #308, opened 2026-07-29
maint: remove pre-commit

Kirill888:
* run ruff checks without extra indirection
* remove pre-commit cache management


.github/workflows/main.yml
```diff
@@ -24,18 +24,17 @@ jobs:
...
 
       - name: Run Checks
         run: |
-          echo "🚀 Linting code: Running pre-commit"
-          uv run pre-commit run -a
+          echo "🚀 Linting code: Running ruff format"
+          uv run ruff format --diff
+          
+          echo "🚀 Linting code: Running ruff check"
+          uv run ruff check --fix --show-fixes --output-format grouped
```
pjonsson, line 36: Don't you want `--output-format github` here? Will emit something like
```
::error title=ruff (I001),file=/home/ubuntu/odc-geo/odc/geo/converters.py,line=5,endLine=12::odc/geo/converters.py:5:1: I001 Import block is un-sorted or un-formatted%0A  help: Organize imports
```
Which I think will put a dialog box on the faulty line in the PR.

And if you update to Ruff 0.16 in `uv.lock`, you should be able to use the github output format for `ruff format` as well,

.github/workflows/main.yml
```diff
@@ -24,18 +24,17 @@ jobs:
...
-          key: pre-commit-${{ hashFiles('.pre-commit-config.yaml') }}
-
       - name: Set up the environment
         uses: ./.github/actions/setup-python-env
 
       - name: Run Checks
         run: |
-          echo "🚀 Linting code: Running pre-commit"
-          uv run pre-commit run -a
+          echo "🚀 Linting code: Running ruff format"
```
pjonsson, line 32: There should probably be a `uv lock --check` first here to ensure that the tools used are consistent with the dev group in `pyproject.toml`.

pjonsson (approved):
See comments about Ruff 0.16, checking `uv.lock`, and using `--output-format github`. I'll approve now so you can merge when I'm asleep.

.github/workflows/main.yml
Kirill888: I just copied what pre-commit was running

.github/workflows/main.yml
Kirill888: I don't know, seems like that sort of thing should be part of setup env, probably is even, or a separate step above this one.

.github/workflows/main.yml
Kirill888: these can be addressed later on when we consolidate checks into a shared action.

PR #310, opened 2026-07-31
fix: densify measured |p1|² + |p2|² instead of the segment distance

gaoflow:
densify's `short_enough` predicate measured `p1[0]**2 + p2[0]**2` instead of the squared segment distance, so long segments whose endpoints sit near zero were treated as already fine and never densified. It now measures the actual segment. Empty and non-positive-`resolution` inputs are guarded (they used to spin forever), and `resolution="auto"` falls back to the extent for zero-area geometries. This is the densify part of #306, split out as discussed.

### DIFF
--- a/odc/geo/geom.py
+++ b/odc/geo/geom.py
@@ -458,10 +458,14 @@ def densify(
     """
     Adds points so they are at most `resolution` units apart.
     """
+    if len(coords) == 0:
+        return []
+    assert resolution > 0
+
     d2 = resolution**2
 
     def short_enough(p1, p2):
-        return (p1[0] ** 2 + p2[0] ** 2) < d2
+        return ((p1[0] - p2[0]) ** 2 + (p1[1] - p2[1]) ** 2) < d2
 
     new_coords: List[Tuple[float, float]] = [cast(Tuple[float, float], coords[0])]
     for p1, p2 in zip(coords[:-1], coords[1:]):
@@ -1532,4 +1536,9 @@ def count_coordinates(g: Geometry | Iterable[Geometry] | BoundingBox) -> int:
 
 def _auto_resolution(g: Geometry) -> float:
     # aim for ~100 points per side of a square
-    return math.sqrt(g.area) * 4 / 100
+    if g.area > 0:
+        return math.sqrt(g.area) * 4 / 100
+    # zero-area inputs (lines, degenerate polygons) still need densifying,
+    # fall back to the extent
+    bbox = g.boundingbox
+    return max(bbox.span_x, bbox.span_y) / 25
--- a/odc/geo/geom.py
+++ b/odc/geo/geom.py
@@ -460,7 +460,13 @@ def densify(
     """
     if len(coords) == 0:
         return []
-    assert resolution > 0
+    if resolution <= 0:
+        # A degenerate (zero-extent) geometry legitimately produces resolution=0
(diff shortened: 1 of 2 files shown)


odc/geo/geom.py
```diff
@@ -1532,4 +1538,10 @@ def count_coordinates(g: Geometry | Iterable[Geometry] | BoundingBox) -> int:
 
 def _auto_resolution(g: Geometry) -> float:
     # aim for ~100 points per side of a square
-    return math.sqrt(g.area) * 4 / 100
+    res = math.sqrt(g.area) * 4 / 100
+    if res > 0:
```
pjonsson: https://github.com/opendatacube/odc-geo/pull/306#discussion_r3655424549

odc/geo/geom.py
```diff
@@ -458,10 +458,16 @@ def densify(
     """
     Adds points so they are at most `resolution` units apart.
     """
+    if len(coords) == 0:
+        return []
+    if not resolution > 0:
```
pjonsson: https://github.com/opendatacube/odc-geo/pull/306#discussion_r3655583570

binggao1230:
Addressed both linked review points in 6f3674c: `_auto_resolution` now directly branches on `g.area > 0`, and `densify` asserts a positive resolution rather than returning a special-case result. Targeted geometry tests pass (56 passed, 4 skipped, 1 xfailed), with ruff clean.

Kirill888:
Tests are failing due to added assert, it is triggered with `resolution=0` computed here:

https://github.com/opendatacube/odc-geo/blob/95accac4fa6858ee91f44cb27d1728e9a9aa262e/odc/geo/geobox.py#L250-L253

basically `resolution=0` can be supplied directly to `Geometry.to_crs()` method also, not just computed by the auto method.

For a robust implementation we should check for `resolution=0` around these parts:

https://github.com/opendatacube/odc-geo/blob/95accac4fa6858ee91f44cb27d1728e9a9aa262e/odc/geo/geom.py#L755-L761

and skip adding extra segments, I say we should treat `resolution=0` same as `resolution=None`, not sure if that should be silent or generate a warning.

odc/geo/geom.py
```diff
@@ -1532,4 +1536,9 @@ def count_coordinates(g: Geometry | Iterable[Geometry] | BoundingBox) -> int:
 
 def _auto_resolution(g: Geometry) -> float:
     # aim for ~100 points per side of a square
-    return math.sqrt(g.area) * 4 / 100
+    if g.area > 0:
+        return math.sqrt(g.area) * 4 / 100
+    # zero-area inputs (lines, degenerate polygons) still need densifying,
+    # fall back to the extent
+    bbox = g.boundingbox
+    return max(bbox.span_x, bbox.span_y) / 25
```
Kirill888, line 1553: this will still be zero for single point `g`

binggao1230:
Confirmed and fixed in 2c7aebf — a single-point/empty-shape `g` does still hit `resolution=0` (`_reproject_resolution`, `outline()` in ui.py all derive it from a zero-span bbox), and that's exactly what `test_html_repr` was hitting on an empty GeoBox slice. Since there's no risk of the loop running forever when `resolution<=0` (nothing to subdivide), `densify()` now just returns the coordinates unchanged in that case instead of asserting. Added a regression test exercising the zero-shape GeoBox path directly. All local tests green.

Kirill888 (approved):

